PR #814, opened 2025-01-27
Adikov/reporter second signature

melatron:


apps/trigger-oracle/src/lib.rs
```diff
@@ -473,21 +507,129 @@ impl OracleTrigger {
...
+        kafka_report_endpoint: Option<String>,
+        signal_sender: UnboundedSender<ConsensusSecondRoundBatch>,
+        _kafka_info: Option<String>,
+    ) -> TerminationReason {
+        // TODO(adikov): remove unwrap/expect
+        // TODO(adikov): get all kafka configuration from `kafka_info` parameter
+
+        let Some(kafka_report_endpoint) = kafka_report_endpoint else {
+            return TerminationReason::Other("No kafka report provided".to_string());
+        };
```
Xearty, line 534: The error message is misleading. Maybe "No kafka endpoint provided" is more appropriate.

apps/trigger-oracle/src/lib.rs
```diff
@@ -473,21 +496,98 @@ impl OracleTrigger {
...
+                    Err(err) => {
+                        // Handle message errors
+                        tracing::error!("Error while consuming: {:?}", err);
+                    }
+                }
+            }
+        }
+
+        TerminationReason::Other("Signal secondary consensus loop terminated".to_string())
+    }
```
Xearty: Is there a reason for not using `while let`?

Xearty: What should be the expected behavior if the kafka server exits? Should the task keep going or should it halt? We should test some recovery strategy if the kafka server stops. We should probably sleep to avoid busy loops if we get None and try to reconnect.

apps/trigger-oracle/src/lib.rs
```diff
@@ -473,21 +507,129 @@ impl OracleTrigger {
...
```
Xearty, line 602: Is there a reason for not using `while let`?

Xearty: What should be the expected behavior if the kafka server exits? Should the task keep going or should it halt? We should test some recovery strategy if the kafka server stops. We should probably sleep to avoid busy loops if we get None and try to reconnect.

libs/gnosis_safe/src/utils.rs
```diff
@@ -97,6 +97,37 @@ pub fn create_private_key_signer(private_key: &str) -> PrivateKeySigner {
     .unwrap()
 }
 
+fn hex_string_to_bytes(hex_string: &str) -> Result<[u8; 32], String> {
+    let hex_string = if hex_string.starts_with("0x") {
+        &hex_string[2..] // Slice to remove "0x"
+    } else {
+        hex_string
+    };
```
Xearty, line 103:
```rs
let hex_string = hex_string.strip_prefix("0x").unwrap_or(&hex_string);
```
Let's make this a reusable function

libs/gnosis_safe/src/utils.rs
```diff
@@ -95,6 +95,37 @@ pub fn create_private_key_signer(private_key: &str) -> PrivateKeySigner {
     .unwrap()
 }
 
+fn hex_string_to_bytes(hex_string: &str) -> Result<[u8; 32], String> {
```
Xearty, line 98: The name suggests that any hex string is allowed and doesn't imply fixed size. Maybe name this `hex_str_to_bytes32` and make it return a `FixedBytes<32>`

libs/gnosis_safe/src/utils.rs
```diff
@@ -95,6 +95,37 @@ pub fn create_private_key_signer(private_key: &str) -> PrivateKeySigner {
...
+        bytes[i / 2] = byte; // Index into the array
+    }
+
+    Ok(bytes)
+}
+
+pub fn create_fixed_bytes(tx: &str) -> FixedBytes<32> {
+    let tx = hex_string_to_bytes(tx).unwrap();
+    FixedBytes::<32>::new(tx)
+}
```
Xearty, line 127: I don't see much point in this function

libs/gnosis_safe/src/utils.rs
```diff
@@ -144,6 +175,16 @@ pub fn signature_to_bytes(signature: PrimitiveSignature) -> Vec<u8> {
...
+pub fn bytes_to_hex_string(bytes: Vec<u8>) -> String {
+    // Changed to Vec<u8>
+    let mut hex_string = String::new();
+    for byte in bytes {
+        // Iterate directly over the Vec
+        hex_string.push_str(&format!("{:02x}", byte));
+    }
+    hex_string
+}
+
```
Xearty, line 187: The function doesn't require owned bytes. Also you could use `hex::encode` instead of this function

apps/trigger-oracle/src/lib.rs
```diff
@@ -473,21 +507,129 @@ impl OracleTrigger {
...
             .name("sender to sequencer")
             .spawn(future)
             .expect("sender to sequencer failed to start")
     }
 
+    async fn signal_secondary_signature(
+        kafka_report_endpoint: Option<String>,
+        signal_sender: UnboundedSender<ConsensusSecondRoundBatch>,
+        _kafka_info: Option<String>,
+    ) -> TerminationReason {
```
Xearty, line 528: What will `kafka_info` contain?

apps/trigger-oracle/src/lib.rs
```diff
@@ -473,21 +507,129 @@ impl OracleTrigger {
...
+        }
+
+        TerminationReason::Other("Signal secondary consensus loop terminated".to_string())
+    }
+
+    fn start_manager(
+        rx: UnboundedReceiver<(String, Payload)>,
+        ss_rx: UnboundedReceiver<ConsensusSecondRoundBatch>,
+        sequencer_post_batch: &url::Url,
+        sequencer_aggregated_consensus: &url::Url,
```
Xearty, line 608: Please name the urls like `.*_url`

apps/trigger-oracle/src/lib.rs
```diff
@@ -473,21 +507,129 @@ impl OracleTrigger {
...
+        secret_key: &str,
+        second_consensus_secret_key: &str,
+        reporter_id: u64,
+    ) -> Vec<tokio::task::JoinHandle<TerminationReason>> {
```
Xearty, line 612: Import the symbol `JoinHandle` so you don't have to fully qualify it

apps/trigger-oracle/src/lib.rs
```diff
@@ -566,6 +708,53 @@ impl OracleTrigger {
         TerminationReason::SequencerExitRequested
     }
 
+    async fn process_aggregated_consensus(
+        mut ss_rx: UnboundedReceiver<ConsensusSecondRoundBatch>,
+        sequencer: url::Url,
```
Xearty, line 713: rename this to `sequencer_url`

Xearty: Fully import `Url`

Xearty (approved):

apps/sequencer/src/providers/eth_send_utils.rs
```diff
@@ -127,14 +127,14 @@ pub async fn get_serialized_updates_for_network(
...
-        return Ok("".to_string());
+        return Ok(Vec::new());
     }
 
     let contract_version = provider
         .get_contract(PRICE_FEED_CONTRACT_NAME)
         .ok_or(eyre!("{PRICE_FEED_CONTRACT_NAME} contract is not set!"))?
         .contract_version;
-    drop(provider);
+    // drop(provider);
```
HristoStaykov, line 137: Do we need to keep the lock on provider? If we do, we should remove the debug log that states we are releasing it.

apps/sequencer/src/providers/eth_send_utils.rs
```diff
@@ -155,17 +155,14 @@ pub async fn get_serialized_updates_for_network(
             *feed_id,
             FeedStrideAndDecimals::from_feed_config(&feed_config),
         );
-        drop(feed_config);
```
HristoStaykov, line 158: Also for the feed_config, might be we want to release the lock implicitly as soon as we are done using the object?

libs/feeds_processing/src/adfs_gen_calldata.rs
```diff
@@ -311,23 +311,27 @@ pub mod tests {
...
+            hex::encode(
+                adfs_serialize_updates(
+                    net,
+                    &updates,
+                    Some(&round_counters),
+                    config.clone(),
+                    &mut feeds_rounds,
+                )
+                .await
+                .unwrap()
```
HristoStaykov, line 323: Might be an expect will be better option, even though the conversion should never fail.

libs/feeds_processing/src/adfs_gen_calldata.rs
```diff
@@ -311,23 +311,27 @@ pub mod tests {
...
         // Call as it will be in the reporter (feeds_rounds provided by the sequencer)
         assert_eq!(
             expected_result,
-            adfs_serialize_updates(net, &updates, None, config, &mut feeds_rounds,)
-                .await
-                .unwrap()
+            hex::encode(
+                adfs_serialize_updates(net, &updates, None, config, &mut feeds_rounds,)
+                    .await
+                    .unwrap()
```
HristoStaykov, line 333: Same as above.

libs/feeds_processing/src/utils.rs
```diff
@@ -251,13 +252,15 @@ pub async fn perform_anomaly_detection(
         .context("Failed to join feed slots manager anomaly detection!")?
 }
 
-pub async fn validate(
-    feeds_config: HashMap<u32, FeedStrideAndDecimals>,
-    mut batch: ConsensusSecondRoundBatch,
-    last_votes: HashMap<u32, VotedFeedUpdate>,
-    tolerated_deviations: HashMap<u32, f64>,
+fn check_aggregated_votes_deviation(
```
HristoStaykov, line 255: Good to have this method renamed, this way it is more search friendly :)

nix/modules/sequencer/sequencer-opts.nix
```diff
@@ -79,7 +79,7 @@ let
 
       safe_min_quorum = mkOption {
         type = types.int;
-        default = 5;
+        default = 1;
```
HristoStaykov, line 82: A reminder for near future - we might want to change this parameter to be the percentage of reporters that form a quorum.

HristoStaykov (approved):

apps/sequencer/src/providers/eth_send_utils.rs
Xearty: What do you mean by releasing the lock implicitly. The drop function is being called implicitly at the end of the scope

libs/feeds_processing/src/adfs_gen_calldata.rs
Xearty: It was unwrap already. I haven't changed this line

apps/sequencer/src/providers/eth_send_utils.rs
HristoStaykov: I missed that there is nothing else in the scope afterwards. It is ok to remove the drop here.

PR #953, opened 2025-02-21
Hstaykov/feed processing to libs

HristoStaykov:
Implement the necessary checks a Reporter needs to perform before participating in the second round of consensus. The sequencer provides all the votes from a quorum of reporters + their signatures that justify an aggregate update. The checks are in libs/feeds_processing/src/utils.rs:validate function.

### DIFF
--- a/apps/sequencer/src/feeds/feeds_slots_manager.rs
+++ b/apps/sequencer/src/feeds/feeds_slots_manager.rs
@@ -270,7 +270,7 @@ pub async fn register_feed_with_config(
         }
 
         let new_feed_metadata = FeedMetaData::new(
-            &new_name,
+            new_name,
             new_feed_config.report_interval_ms,
             new_feed_config.quorum_percentage,
             new_feed_config.skip_publish_if_less_then_percentage,
--- a/libs/data_feeds/src/feeds_processing.rs
+++ b/libs/data_feeds/src/feeds_processing.rs
@@ -85,8 +85,6 @@ impl VotedFeedUpdate {
     }
 }
 
-pub const REPORT_HEX_SIZE: usize = 64;
-
 pub fn naive_packing(feed_result: &FeedType, digits_in_fraction: usize) -> Vec<u8> {
     //TODO: Return Bytes32 type
     feed_result.as_bytes(digits_in_fraction)
--- a/libs/data_feeds/src/feeds_processing.rs
+++ b/libs/data_feeds/src/feeds_processing.rs
@@ -93,14 +93,14 @@ pub fn naive_packing(feed_result: &FeedType, digits_in_fraction: usize) -> Vec<u
 use std::collections::HashMap;
 
 #[derive(Debug, Clone, Default)]
-pub struct UpdateToSend {
+pub struct BatchedAggegratesToSend {
     pub block_height: u64,
     pub updates: Vec<VotedFeedUpdate>,
     // The key in this map is the feed id for which we provide a proof for the aggregated value.
     pub proofs: HashMap<u32, Vec<DataFeedPayload>>,
 }
 
-impl UpdateToSend {
+impl BatchedAggegratesToSend {
     // The updates to be sent to different networks go through multiple filters.
     // Eventually we might need to reduce the proof to only contain records for
     // the relevant updates. The following function does that and returns the
--- a/libs/data_feeds/src/feeds_processing.rs
+++ b/libs/data_feeds/src/feeds_processing.rs
@@ -92,6 +92,39 @@ pub fn naive_packing(feed_result: &FeedType, digits_in_fraction: usize) -> Vec<u
     feed_result.as_bytes(digits_in_fraction)
 }
 
+use std::collections::HashMap;
+
+#[derive(Debug, Clone, Default)]
+pub struct UpdateToSend {
+    pub block_height: u64,
+    pub updates: Vec<VotedFeedUpdate>,
+    // The key in this map is the feed id for which we provide a proof for the aggregated value.
+    pub proofs: HashMap<u32, Vec<DataFeedPayload>>,
+}
+
+impl UpdateToSend {
+    // The updates to be sent to different networks go through multiple filters.
+    // Eventually we might need to reduce the proof to only contain records for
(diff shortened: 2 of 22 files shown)


libs/feeds_processing/src/utils.rs
```diff
@@ -0,0 +1,612 @@
...
+        .context("Failed to spawn feed slots manager anomaly detection!")?
+        .await
+        .context("Failed to join feed slots manager anomaly detection!")?
+}
+
+type FeedId = u32;
+type FeedTimestamp = u64;
+type FeedIdToTimetamp = HashMap<FeedId, FeedTimestamp>;
+
+pub async fn validate_sigcheck(
```
smanilov, line 254: This function is not used anywhere.

Has it been tested in any way?

Same for validate.

libs/feed_registry/src/types.rs
```diff
@@ -71,7 +72,7 @@ impl FeedMetaData {
 
     #[allow(clippy::too_many_arguments)]
     pub fn new(
-        name: &str,
+        name: String,
```
smanilov, line 75: Why this change? It seems completely unnecessary, given that `.to_string()` is called on `name` again.

apps/sequencer/src/feeds/feeds_slots_manager.rs
```diff
@@ -270,7 +270,7 @@ pub async fn register_feed_with_config(
         }
 
         let new_feed_metadata = FeedMetaData::new(
-            &new_name,
```
smanilov, line 273: This is the only place where a reference is taken, so perhaps just avoid the clone a few lines earlier?

libs/feeds_processing/src/adfs_gen_calldata.rs
```diff
@@ -63,30 +66,15 @@ pub async fn adfs_serialize_updates(
...
-        );
 
-        let stride = match &feed_config {
-            Some(f) => f.stride,
+        let (stride, digits_in_fraction) = match &feeds_config.get(&feed_id) {
+            Some(f) => (f.stride, f.decimals),
             None => {
-                warn!("Propagating result for unregistered feed! Support left for legacy one shot feeds of 32 bytes size.");
-                1
+                error!("Propagating result for unregistered feed! Support left for legacy one shot feeds of 32 bytes size. Decimale default to 18");
```
smanilov, line 73: typo: Decimale should be Decimal.

libs/feeds_processing/src/adfs_gen_calldata.rs
```diff
@@ -48,7 +47,11 @@ pub async fn adfs_serialize_updates(
     net: &str,
     feed_updates: &BatchedAggegratesToSend,
     feeds_metrics: Option<Arc<RwLock<FeedsMetrics>>>,
-    feeds_config: Arc<RwLock<HashMap<u32, FeedConfig>>>,
+    feeds_config: HashMap<u32, FeedStrideAndDecimals>,
+    feeds_rounds: &mut HashMap<u32, u64>, /* The rounds table for the relevant feeds. If the feeds_metrics are provided,
+                                          this map will be filled with the update count for each feed from it. If the
+                                          feeds_metrics is None, feeds_rounds will be used as the source of the updates
+                                          count. */
```
smanilov, line 54: In terms of implementation this is good. In terms of design, I'm not so sure.

We now have an argument, that depending on another argument is sometimes an input parameter and sometimes an output parameter.

This is a rather complicated design in my opinion. I don't have a better suggestion, though.

libs/feeds_processing/src/adfs_gen_calldata.rs
```diff
@@ -48,7 +47,11 @@ pub async fn adfs_serialize_updates(
     net: &str,
     feed_updates: &BatchedAggegratesToSend,
     feeds_metrics: Option<Arc<RwLock<FeedsMetrics>>>,
-    feeds_config: Arc<RwLock<HashMap<u32, FeedConfig>>>,
+    feeds_config: HashMap<u32, FeedStrideAndDecimals>,
```
smanilov, line 50: Perhaps rename parameter to `strides_and_decimals`, as it's called elsewhere. `feeds_config` is misleading, since there's a type with the same name.

smanilov (approved):
Looks good if you address the comments.

How did you test `validate` and `validate_sigcheck`?

libs/data_feeds/src/feeds_processing.rs
```diff
@@ -85,13 +85,43 @@ impl VotedFeedUpdate {
...
+        if removed_elements_count > 0 {
+            let mut normalized_proofs = HashMap::new();
+            for update in self.updates.iter() {
+                let Some(proof) = self.proofs.remove(&update.feed_id) else {
+                    error!("Logical ERROR! no proof found for key: {}", update.feed_id);
+                    continue;
+                };
+                normalized_proofs.insert(update.feed_id, proof);
+            }
+        }
```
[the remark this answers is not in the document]
HristoStaykov: Indeed, this was a reminder that was resolved, thanks.

libs/feed_registry/src/types.rs
HristoStaykov: I have removed the to_string call.

libs/feeds_processing/src/adfs_gen_calldata.rs
HristoStaykov: I agree that it is a bit complicated, but we need to be able to call this function both from within the sequencer (which manages the feed round buffer offsets) and from within the reporter (which receives those offsets from the sequencer and uses the provided values as input).

smanilov (approved):

Xearty:
@smanilov Some of the comments have been resolved in https://github.com/blocksense-network/blocksense/pull/980

libs/feed_registry/src/types.rs
Xearty: It's better for a function to take ownership of a String if it needs an owned String in its body

smanilov:
> @smanilov Some of the comments have been resolved in #980

Sweet! Which ones?

PR #973, opened 2025-02-26
Provide timestamp in asset feeds

HristoStaykov:


libs/data_feeds/src/feeds_processing.rs
```diff
@@ -23,10 +23,10 @@ pub struct VotedFeedUpdateWithProof {
 }
 
 impl VotedFeedUpdate {
-    pub fn encode(&self, digits_in_fraction: usize) -> (Vec<u8>, Vec<u8>) {
+    pub fn encode(&self, digits_in_fraction: usize, time_stamp: u64) -> (Vec<u8>, Vec<u8>) {
```
melatron, line 26:
```suggestion
    pub fn encode(&self, digits_in_fraction: usize, timestamp: u64) -> (Vec<u8>, Vec<u8>) {
```
and everywhere else

melatron (approved):
LGTM 
nit: move 18 to a constant.

melatron (approved):

PR #1006, opened 2025-03-07
Integrate new feeds config schema in reporter and sequencer and fix tests

Xearty:


PetarKirov (changes requested):
I will test deploy this branch on `sequencer-001`, but for now let's hold off merging this until we know that the deployment is successful, to keep production environment stable.

libs/feed_registry/src/registry.rs
```diff
@@ -71,7 +72,7 @@ pub fn new_feeds_meta_data_reg_with_test_data() -> FeedMetaDataRegistry {
         always_publish_heartbeat_ms,
         start,
         "Numerical".to_string(),
```
melatron, line 74: Shouldn't it be "numerical"?

melatron (approved):

libs/feed_registry/src/registry.rs
Xearty: Right, we have let this slip by. This field should be refactored to be an enumeration. There are so much occurrences of these literals all over the code. It's a hell to refactor

apps/sequencer/src/http_handlers/admin.rs
```diff
@@ -497,6 +497,29 @@ pub async fn get_oracle_scripts(
                 ],
                 capabilities: HashSet::from_iter(vec![]),
             },
+            OracleScript {
```
yordanmadzhunkov, line 500: Finally yeah !

melatron (dismissed):

PetarKirov (dismissed):
LFG

PetarKirov (approved):

PR #1040, opened 2025-03-13
feat(sequencer): Add get_history endpoint

smanilov:
Tested with:

```
cargo test && cd apps/sequencer && cargo test && cd ../..
```

Also,

```
direnv reload
sed -i 's|CoinMarketCap|crypto-price-feeds|g' config/feeds_config.json
process-compose
```

and in a separate terminal:

```
curl -i -X GET 127.0.0.1:5553/get_history
```


apps/sequencer_tests/bin/sequencer_tests.rs
```diff
@@ -548,6 +549,49 @@ async fn main() -> Result<()> {
         assert_eq!(expected_response, actual_response);
     }
 
+    fn mask_timestamps(response: &str) -> String {
```
yordanmadzhunkov, line 552: Interesting technique, seems like I like it. Today I learned something new .. yeah !

yordanmadzhunkov (approved):

Xearty:
LGTM

Xearty (approved):

apps/sequencer_tests/bin/sequencer_tests.rs
smanilov: Thanks, it's a bit of a hack, but I thought it's good enough for this test :-)

Xearty (commented):
Forgot to submit this review :face_exhaling:

PR #1067, opened 2025-03-17
WIP [docs.ws]: Add common `TextArea` component

kalinsky-dev:


apps/docs.blocksense.network/components/common/TextArea.tsx
```diff
@@ -0,0 +1,58 @@
...
+        </label>
+      )}
+      <textarea
+        id={name}
+        name={name}
+        ref={textareaRef}
+        aria-label={label || 'TextArea'}
+        aria-describedby={name ? `${name}-description` : undefined}
+        className={cn(
+          'textarea__content text-sm flex min-h-[80px] w-full px-3 py-2 rounded-md shadow-md border border-solid border-neutral-200 bg-white dark:bg-neutral-900 dark:border-neutral-600 disabled:cursor-not-allowed disabled:opacity-50',
```
VanyaPavlova, line 51: Can we inherit the states (hover, focus) from [Input](https://github.com/blocksense-network/blocksense/blob/main/libs/ts/ui/src/components/Input/Input.tsx) component, as part of Blocksense ui library? 

In this way, we will have them defined in the same way, later they will be refined by design team.

apps/docs.blocksense.network/components/common/TextArea.tsx
```diff
@@ -0,0 +1,112 @@
...
+        {...props}
+      />
+      {error && (
+        <p className="textarea__error text-sm text-red-500 mt-1">
+          {errorMessage}
+        </p>
+      )}
+    </div>
+  );
+};
```
VanyaPavlova, line 112: Add displayName for the component.

`TextArea.displayName = 'TextArea';`

VanyaPavlova (approved):

PR #1070, opened 2025-03-17
feat(sequencer/http_handlers): Make http handlers' input buffer size configurable

HristoStaykov:


EmilIvanichkovv (approved):
LGTM

Tested by setting different values of `http_input_buffer_size` in the `process-compose` set up and it worked as expected

PR #1073, opened 2025-03-18
fix(deployment): Name of gecko oracle script

smanilov:
@PetarKirov mentioned there might be more changes necessary.

Also, pre-commit for clippy seems to be broken, so I committed with `--no-verify`:

```
clippy...................................................................Failed                                       
- hook id: clippy                                                                                                     
- exit code: 101                                                                                                      
                                                                                                                      
error: failed to select a version for the requirement `json-patch = "^4.0.0"` (locked to 4.0.0)                       
candidate versions found which didn't match: 3.0.1                                                                    
location searched: crates.io index                                                                                    
required by package `sequencer_tests v0.1.1 (/home/stan/code/repos/blocksense/apps/sequencer_tests)`                  
As a reminder, you're using offline mode (--offline) which can sometimes cause surprising resolution failures, if this
 error is too confusing you may wish to retry without `--offline`.
```


apps/sequencer/src/http_handlers/admin.rs
```diff
@@ -511,7 +511,7 @@ pub async fn get_oracle_scripts(
                 capabilities: HashSet::new(),
             },
             OracleScript {
-                id: "gecko_terminal".to_string(),
+                id: "gecko-terminal".to_string(),
```
smanilov, line 514: This name needs to match contents of feeds_config:

```
      "script": "gecko-terminal",
```

EmilIvanichkovv:
> Also, pre-commit for clippy seems to be broken

I had a problem with `pre-commit` that got resolved after `cargo clean`

PR #1076, opened 2025-03-18
[config]: Add `ExSat BTC`, `USDT / WMON 0.2%` & pegged to 1 `USDC / USD` & `USDT / USD`

EmilIvanichkovv:
To test if all works I start `process-compose` and try to read the on-chain data

#### Read data for `USDT / USD Pegged` with id `50000`
```
cast call 0xee5a4826068c5326a7f06fd6c7cbf816f096846c --data 0x8000c350 --rpc-url http://127.0.0.1:8546 |  cut -c1-50 | cast to-dec
```

#### Read data for `USDC / USD Pegged` with id `50001`
```
cast call 0xee5a4826068c5326a7f06fd6c7cbf816f096846c --data 0x8000c351 --rpc-url http://127.0.0.1:8546 |  cut -c1-50 | cast to-dec
```

#### Read data for `ExSat BTC` with id `100000`
```
cast call 0xee5a4826068c5326a7f06fd6c7cbf816f096846c --data 0x800186a0 --rpc-url http://127.0.0.1:8546 |  cut -c1-50 | cast to-dec
```

```
 cast call 0xee5a4826068c5326a7f06fd6c7cbf816f096846c --data 0x800f4240 --rpc-url http://127.0.0.1:8546 |  cut -c1-50 | cast to-dec     
```

PR #1078, opened 2025-03-18
fix[oracles/gecko-terminal]: Fix settings parsing

EmilIvanichkovv:
## Key changes

- Updated `spin.toml` format to align with `feeds_config_v2`.  
- Fixed parsing of feed settings to support shared configuration files in local and production environments.  
- Introduced a temporary check for `"pool"` in the data feed to differentiate `gecko-terminal` feeds and prevent deserialization errors.  
- Implemented `get_resources_from_settings` to safely extract relevant arguments.  

## More details

One part of the changes is ensuring that the oracle works with the updated version of the data in `spin.toml`, which is generated from `feeds_config_v2`.

However, there is a more complex issue. The check `if feed_setting.data.contains("pool")` is a temporary solution to address the following problem:

When running the oracle locally with the command `just start-oracle gecko-terminal`, the `apps/oracles/gecko-terminal/spin.toml` file is read, and the workflow functions as expected. But when running oracle with our `process-compose` or even in production, we do not have separate `spin.toml` files for each oracle. Instead, all feed data is stored in the same file (in local test environments, this is `.devenv/state/blocksense/reporter/a/spin.toml`). As a result, the previous approach of deserializing the settings led to errors like "Can't parse GeckoTerminalPool" because most feeds couldn't process the `arguments` deserializing due to mismatched schemas.

To solve this, we now check for the string "pool" in the feed data, which is unique to the `gecko-terminal` oracle. This allows us to safely decode the arguments.

As mentioned, this is a temporary solution. We should eventually implement a way to differentiate feed data based on the oracles that handle them. This was already communicated ( or soon will be ) with team members.

cc: @PetarKirov 

## Future tasks
- [ ]  Properly differentiate feeds is `spin.toml` based on their respective oracles.

## How to test
- Localy run following command and monitor logs:
```
just start-oracle gecko-terminal  
```
- You can make e2e test with `process-compose` once https://github.com/blocksense-network/blocksense/pull/1076 is merged. Follow this steps
```bash
direnv reload
process-compose
```
Examine logs of the reporter
To check the on-chain data you can run 
```
 cast call 0xee5a4826068c5326a7f06fd6c7cbf816f096846c --data 0x800f4240 --rpc-url http://127.0.0.1:8546 |  cut -c1-50 | cast to-dec     
 ```


yordanmadzhunkov (dismissed):
Awesome!

PetarKirov (approved):

PR #1080, opened 2025-03-19
Remove alloy-encode-packed and upgrade alloy to 0.11

smanilov:
Most of the changed lines are in Cargo.lock. Actual change is only 102 lines (added/removed combined).

The dependency on `alloy-encode-packed` is removed, since it forces a transitive dependency on alloy 0.7.

The code from it is copied in and minimized to only the part that we're using.

Added a new field in Provider -- rpc_url -- to make it easy to extract the Url of an RPC provider. This is needed due to a change in the alloy API from 0.9 to 0.11 which hides the type of the transport (Http). An alternative would be to use dynamic casting, but I decided on this solution.

Cannot upgrade to alloy 0.12 due to following error in a dependency of alloy:

```
error[E0599]: no function or associated item named `hoodi` found for struct `alloy_chains::Chain` in the current scope
   --> /home/stan/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/alloy-hardforks-0.1.1/src/hardfork/ethereum.rs:63:28
    |
63  |         if chain == Chain::hoodi() {
    |                            ^^^^^ function or associated item not found in `Chain`
    |
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -90,7 +90,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
  "quote",
- "syn 2.0.96",
+ "syn 2.0.100",
 ]
 
 [[package]]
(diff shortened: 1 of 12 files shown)


RadoslavDimchev (approved):
Sweet

PR #1089, opened 2025-03-19
chore: Improve Shell Script Robustness

zkpepe:
### **Description:**  
I noticed some potential pitfalls in the script and made a few fixes to improve its reliability:  

- Wrapped all variable expansions in quotes (`"$VAR"`) to prevent issues with spaces in paths.  
- Ensured `${HASH}` is always treated as a string in JSON—this prevents syntax errors when empty or containing unexpected characters.  
- Used `"$DIR/trigger-oracle.json"` instead of `$DIR/trigger-oracle.json` for the same reason.  
- Added a check to verify if the binary exists before running `tar`, preventing misleading errors.  

 🚀


PetarKirov:
The file you're improving is actually obsolete, so we will remove it instead - see #1090.

Thanks anyway for your interest in contributing, @zkpepe!

PR #1091, opened 2025-03-20
WIP [docs.ws]: Add common `RadioGroup` component

kalinsky-dev:


apps/docs.blocksense.network/components/common/RadioGroup.tsx
```diff
@@ -0,0 +1,121 @@
...
+        type="radio"
+        id={value}
+        name={name}
+        value={value}
+        checked={isSelected}
+        onChange={handleChange}
+        className="radio-group__input"
+      />
+      <label htmlFor={value} className="radio-group__label cursor-pointer">
+        {children}
```
VanyaPavlova, line 113: Let' use our Label component as part of this [PR](https://github.com/blocksense-network/blocksense/pull/938).
If we have controls, that use simple label element, we wil replace it at once into separate item, because it may need other props to set, but for this PR, let's replace it here.

apps/docs.blocksense.network/components/common/RadioGroup.tsx
```diff
@@ -0,0 +1,121 @@
...
+  ariaLabel,
+  children,
+  ...props
+}: RadioGroupProps) => {
+  return (
+    <RadioGroupContext.Provider value={{ value, name, onChange }}>
+      <div
+        {...props}
+        role="radiogroup"
+        aria-label={ariaLabel}
```
VanyaPavlova, line 43: Leaving a note - If we change the div to a fieldset, we won't need to set explicitly role="radiogroup".
More details [here. ](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/radiogroup_role)

apps/docs.blocksense.network/components/common/RadioGroup.tsx
```diff
@@ -0,0 +1,121 @@
...
+  onChange,
+  ariaLabel,
+  children,
+  ...props
+}: RadioGroupProps) => {
+  return (
+    <RadioGroupContext.Provider value={{ value, name, onChange }}>
+      <div
+        {...props}
+        role="radiogroup"
```
VanyaPavlova, line 42: Leaving a note - If we change the div to a fieldset, we won't need to set explicitly role="radiogroup".
More details [here. ](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/radiogroup_role)

apps/docs.blocksense.network/components/common/RadioGroup.tsx
```diff
@@ -0,0 +1,123 @@
...
+  return (
+    <p
+      {...props}
+      onClick={handleChange}
+      className={cn(
+        'radio-group__item flex items-center gap-2 cursor-pointer',
+        props.className,
+      )}
+    >
+      <input
```
RadoslavDimchev, line 105: Here you can use our `Input` component from `@blocksense/ui/Input` by updating its type definitions to fully support default `HTML` input attributes, such as id and name, ensuring they conform to valid HTML standards. Also, remove any redundant props that are provided by default. You can refer to the `TextArea` component as a guide on how we used the correct types.

apps/docs.blocksense.network/components/common/RadioGroup.tsx
```diff
@@ -0,0 +1,123 @@
...
+type RadioGroupLabelProps = HTMLAttributes<HTMLLabelElement> & {
+  children: ReactNode;
+};
+
+export const RadioGroupLabel = ({
+  children,
+  ...props
+}: RadioGroupLabelProps) => {
+  return (
+    <label
```
RadoslavDimchev, line 65: Use our `Label` component as in the `RadioGroupItem`.

RadoslavDimchev (changes requested):

apps/docs.blocksense.network/components/common/RadioGroup.tsx
```diff
@@ -0,0 +1,124 @@
...
+  };
+
+  return (
+    <div
+      {...props}
+      onClick={handleChange}
+      className={cn(
+        'radio-group__item flex items-center gap-2 cursor-pointer',
+        props.className,
+      )}
```
RadoslavDimchev:
```suggestion
      onClick={handleChange}
      className={cn(
        'radio-group__item flex items-center gap-2 cursor-pointer',
        props.className,
      )}
     {...props}
```

Can you pass `{...props}` in the bottom of component props (in all components)

libs/ts/ui/src/components/Input/Input.tsx
```diff
@@ -32,22 +26,11 @@ const sizes = {
   xl: 'h-14 text-xl px-5',
 };
 
-type InputBaseProps = {
-  className?: string;
+type InputBaseProps = Omit<InputHTMLAttributes<HTMLInputElement>, 'size'> & {
```
RadoslavDimchev, line 29:
```suggestion
type InputBaseProps = InputHTMLAttributes<HTMLInputElement & {
```

You can remove `Omit` and rename `size` prop to `inputSize`

apps/docs.blocksense.network/components/common/RadioGroup.tsx
```diff
@@ -0,0 +1,124 @@
...
+};
+
+const RadioGroupContext = createContext<RadioGroupContextValue>(
+  {} as RadioGroupContextValue,
+);
+
+type RadioGroupProps = Omit<
+  FieldsetHTMLAttributes<HTMLFieldSetElement>,
+  'onChange'
+> & {
```
RadoslavDimchev, line 28:
```suggestion
type RadioGroupProps = FieldsetHTMLAttributes<HTMLFieldSetElement> & {
```

You can remove `Omit` and rename `onChange` prop to `onValueChange`.

apps/docs.blocksense.network/components/common/RadioGroup.tsx
```diff
@@ -0,0 +1,124 @@
...
+  children,
+  ...props
+}: RadioGroupProps) => {
+  return (
+    <RadioGroupContext.Provider value={{ value, name, onChange }}>
+      <fieldset
+        {...props}
+        aria-label={ariaLabel}
+        className={cn('radiogroup flex flex-col gap-1', props.className)}
+      >
```
RadoslavDimchev, line 50:
```suggestion
      <fieldset
        aria-label={ariaLabel}
        className={cn('radiogroup flex flex-col gap-1', props.className)}
        {...props}
      >
```

Can you pass `{...props}` at the bottom of component props (in all components)

apps/docs.blocksense.network/components/common/RadioGroup.tsx
RadoslavDimchev: Yes, it will be overridden, but as we talked you can just destructure it in the props and use directly className like in the other components, also you can remove the custom `ariaLabel` prop, use `aria-label` in the parent and just pass it through the `{...props}`.

RadoslavDimchev (approved):
> We should update `Input` - pass default props by {...props} without explicitly and merge input with input base

PR #1093, opened 2025-03-20
fix(sequencer/adfs): Fix to set round indexes correctly in adfs contracts

HristoStaykov:


EmilIvanichkovv (approved):
This is a hot fix. Tomorrow we will refactor and improve it
cc: @HristoStaykov

PR #1095, opened 2025-03-21
Yordan/enable reading history from chain

yordanmadzhunkov:


EmilIvanichkovv (commented):
At this point it is critical to make it work with ADFS contracts. From my point of view it is best to implement the necessary changes for ADFS and then unable this feature .

After a quick call with @yordanmadzhunkov we cleared small part of the details but please sync with @Owliie for the whole picture

PR #1100, opened 2025-03-21
refactor(sequencer/logging): Add debug logs when incrementing and decrementing round indexes

HristoStaykov:


smanilov (approved):
Reviewed offline. Applied suggestions as latest commit.

PR #1121, opened 2025-03-24
feat(metrics): Switch metrics to gauge

smanilov:
Let's do the data aggregation/visualization in the Prometheus UI, not in the metrics themselves.


smanilov:
The PR is smaller than it seems, because it also includes some reformatting.

The actual change is ~25 lines.

HristoStaykov (approved):

PR #1124, opened 2025-03-25
build(docs.ws): Update no-longer used deps 

VanyaPavlova:
After moving everything from shadcn to our own component common folder, we're removing the no-longer used deps.


EmilIvanichkovv (approved):
CI passes so seems like all good

RadoslavDimchev (approved):

PetarKirov:
Nice!

PR #1126, opened 2025-03-25
feat(nix/modules): Implement Blocksense configuration files abstraction in the blocksense module and expose the configuration files directory as a symlink. Refactor module options

Xearty:


nix/modules/sequencer/config-files.nix
```diff
@@ -0,0 +1,51 @@
+{
+  cfg,
+  self,
+  lib,
+}:
+let
+  inherit (self.lib) dashToUnderscore getReporterConfigFilename;
+
+  configJSON2 =
```
MartinNikov, line 9: You can rename this function and eliminate 2 since you have removed the old configJSON function entirely.

nix/modules/sequencer/config-file-submodule.nix
```diff
@@ -0,0 +1,27 @@
+{
+  name,
```
MartinNikov, line 2: Where did `name` come from? Have you forgotten to call it when creating the submodule, or is there a magic argument for that?

Xearty: it's the name of the key in the attribute set

nix/modules/sequencer/config-files.nix
Xearty: Fair. I dislike the name configJSON overall. I'll try to think of a better name.

PetarKirov (commented):
Looking great so far!

nix/modules/sequencer/backends/systemd.nix
```diff
@@ -58,7 +54,11 @@ let
           StateDirectory = serviceName;
           WorkingDirectory = "/var/lib/${serviceName}";
           ExecStartPre = wasmCopyCmd;
-          ExecStart = "${blocksense.program} node build --from ${reportersConfigJSON.${name}} --up";
+          ExecStart =
+            let
+              reporter-config = cfg.config-files.${self.lib.getReporterConfigFilename name};
+            in
+            "${blocksense.program} node build --from ${reporter-config.path} --up";
```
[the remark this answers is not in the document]
Xearty: I wanted to be able to change the config filename in a single place. This way if we decide to name it differently, we would touch just the function and there is no potential for desynchronization in the services. Why are you opposed to this idea?

.gitignore
monyarm, line 1: Why remove the # Logs heading? Other sections of the .gitignore still have their headings.

nix/lib/default.nix
```diff
@@ -1,8 +1,28 @@
-lib: {
-  dashToUnderscore =
-    attrset:
-    lib.mapAttrs' (name: value: {
-      name = lib.replaceStrings [ "-" ] [ "_" ] name;
-      inherit value;
-    }) attrset;
+lib: rec {
+  kebabToSnakeCase = lib.replaceStrings [ "-" ] [ "_" ];
```
monyarm, line 2:
```suggestion
  dashToUnderscore = lib.replaceStrings [ "-" ] [ "_" ];
  kebabToSnakeCase = dashToUnderscore;
```
Given that kebabToSnakeCase simply turns dashes into underscores, i feel this is a more appropriate name, with kebabToSnakeCase instead being an alias.

This relates to my next suggestion, where we reuse dashToUnderscore.

nix/lib/default.nix
```diff
@@ -1,8 +1,28 @@
...
+  dashToUnderscoreRecursive =
+    attrs:
+    let
+      transformAttr =
+        attr:
+        let
+          value = attrs.${attr};
+        in
+        {
+          name = lib.replaceStrings [ "-" ] [ "_" ] attr;
```
monyarm, line 17:
```suggestion
          name = dashToUnderscore attr;
```

We can reuse the existing function, no need to write it out twice.

nix/modules/blocksense/default.nix
```diff
@@ -0,0 +1,49 @@
...
+              ;
+          };
+        in
+        {
+          options.services.blocksense = import ./module-opts specialArgs;
+
+          config.services.blocksense = {
+            sequencer.providers = lib.mkIf (cfg.anvil != { }) (
+              lib.mapAttrs (_name: value: {
+                "url" = lib.mkDefault "http://127.0.0.1:${toString value.port}";
```
monyarm, line 37: Given that the port is a service option, why not make the ip/endpoint one as well, that defaults to http://127.0.0.1?

```suggestion
                "url" = lib.mkDefault "${value.ip}:${toString value.port}";
```

nix/modules/blocksense/module-opts/anvil.nix
```diff
@@ -33,8 +33,9 @@ in
       description = "The fork URL to use for the Anvil instance.";
     };
```
monyarm, line 35: Given that the port is a service option, why not make the ip/endpoint one as well, that defaults to http://127.0.0.1/?

```suggestion

ip = mkOption {
      type = types.str;
      default = "http://127.0.0.1/;
      default = 8544;
      description = "The ip to use for the Anvil instance.";
    };
```

nix/modules/blocksense/default.nix
Xearty: Unrelated to this PR but anvil in our case is always local

nix/modules/blocksense/module-opts/anvil.nix
Xearty: Unrelated to this PR

PetarKirov (approved):

PR #1146, opened 2025-03-27
WIP [docs.ws]: Integrate feeds config v2 and ADFS contracts to ws

EmilIvanichkovv:


PetarKirov (approved):
Great progress so far! There are various things that I'd like to improve, but in the interest of time, I'll go ahead and merge.

* Use Next.js routing instead of "useEffect" to to handle network selection as this breaks the history (go to `/docs/contracts/deployed-contracts`, select network, select feed from the table and try going back - this does not navigate to the correct page) 
* Make "Oracle Configuration" card full-width
* Add "select network" before network selection drop down in the "Network Access" tab

apps/docs.blocksense.network/src/deployed-contracts/utils.ts
```diff
@@ -0,0 +1,23 @@
+import { DeploymentConfigV2 } from '@blocksense/config-types';
+import { DeploymentConfigArray } from './types';
+
+export function prepareDeploymentData(
+  deploymentConfig: DeploymentConfigArray,
+  feedName: string,
+): DeploymentConfigV2[] {
```
PetarKirov, line 7: Would this work?

```suggestion
import { DeploymentConfigV2 } from '@blocksense/config-types';

export function prepareDeploymentData(
  deploymentConfig: DeploymentConfigV2[],
  feedName: string,
): DeploymentConfigV2[] {
```

PR #1148, opened 2025-03-28
Adding skip-publishing metrics

smanilov:
These metrics are only for the per-feed configuration.

We need a separate set for the per-feed + per-network configuration.

PR #1151, opened 2025-03-28
fix(skip-reason): UnexpectedError should be a dont skip

smanilov:
Addressing https://github.com/blocksense-network/blocksense/pull/1148#discussion_r2018798906


melatron (approved):
LGTM

PR #1167, opened 2025-04-02
[docs.ws]: Implement `useHash` hook

RadoslavDimchev:
Custom hook `useHash` to get and change hash in the URL quickly.


kalinsky-dev (approved):
The hash state is always in sync with the URL hash. I've tested it and it works as expected. Nice!

PR #1179, opened 2025-04-07
[docs.ws]: Add common `Carousel` component

kalinsky-dev:
##### `orientation` (optional)
- **Type**: `'horizontal' | 'vertical'`
- **Default**: `'horizontal'`
- **Description**: Determines the scrolling direction of the carousel. Use `'horizontal'` for left-to-right scrolling or `'vertical'` for top-to-bottom scrolling.

##### `className` (optional)
- **Type**: `string`
- **Description**: Additional utility classes for styling the carousel container.

##### `children` (required)
- **Type**: `ReactNode`
- **Description**: The child components (e.g., `CarouselContent`, `CarouselItem`) to be rendered inside the carousel.

##### `...props` (optional)
- **Type**: `HTMLAttributes<HTMLDivElement>`
- **Description**: Any additional props that can be applied to the carousel container.

---

##### `className` (optional)
- **Type**: `string`
- **Description**: Additional utility classes for styling the carousel content container.

##### `children` (required)
- **Type**: `ReactNode`
- **Description**: The child components (e.g., `CarouselItem`) to be rendered inside the carousel content.

##### `...props` (optional)
- **Type**: `HTMLAttributes<HTMLDivElement>`
- **Description**: Any additional props that can be applied to the carousel content container.

---

##### `className` (optional)
- **Type**: `string`
- **Description**: Additional utility classes for styling the carousel item.

##### `children` (required)
- **Type**: `ReactNode`
- **Description**: The content to be displayed inside the carousel item.

##### `...props` (optional)
- **Type**: `HTMLAttributes<HTMLDivElement>`
- **Description**: Any additional props that can be applied to the carousel item.

---

##### `className` (optional)
- **Type**: `string`
- **Description**: Additional utility classes for styling the previous button.

##### `variant` (optional)
- **Type**: `string`
- **Default**: `'outline'`
- **Description**: The variant of the button (e.g., `'outline'`).

##### `size` (optional)
- **Type**: `string`
- **Default**: `'icon'`
- **Description**: The size of the button.

##### `...props` (optional)
- **Type**: `React.ComponentProps<typeof Button>`
- **Description**: Any additional props that can be applied to the previous button.

---

##### `className` (optional)
- **Type**: `string`
- **Description**: Additional utility classes for styling the next button.

##### `variant` (optional)
- **Type**: `string`
- **Default**: `'outline'`
- **Description**: The variant of the button (e.g., `'outline'`).

##### `size` (optional)
- **Type**: `string`
- **Default**: `'icon'`
- **Description**: The size of the button.

##### `...props` (optional)
- **Type**: `React.ComponentProps<typeof Button>`
- **Description**: Any additional props that can be applied to the next button.

---

#### Carousel Horizontal
```tsx
import {
  Carousel,
  CarouselContent,
  CarouselItem,
  CarouselNext,
  CarouselPrevious,
} from '@/components/common/Carousel';

export const CarouselHorizontal = () => {
  return (
    <Carousel>
      <CarouselContent>
        <CarouselItem>
          <div className="flex items-center justify-center w-48 h-48 border border-gray-300">
            1
          </div>
        </CarouselItem>
        <CarouselItem>
          <div className="flex items-center justify-center w-48 h-48 border border-gray-300">
            2
          </div>
        </CarouselItem>
        <CarouselItem>
          <div className="flex items-center justify-center w-48 h-48 border border-gray-300">
            3
          </div>
        </CarouselItem>
      </CarouselContent>
      <CarouselPrevious />
      <CarouselNext />
    </Carousel>
  );
};
```
#### Carousel Vertical
```tsx
import {
  Carousel,
  CarouselContent,
  CarouselItem,
  CarouselNext,
  CarouselPrevious,
} from '@/components/common/Carousel';

export const CarouselVertical = () => {
  return (
    <Carousel orientation="vertical">
      <CarouselContent className="h-[400px]">
        <CarouselItem>
          <div className="flex items-center justify-center w-48 h-48 border border-gray-300">
            1
          </div>
        </CarouselItem>
        <CarouselItem>
          <div className="flex items-center justify-center w-48 h-48 border border-gray-300">
            2
          </div>
        </CarouselItem>
        <CarouselItem>
          <div className="flex items-center justify-center w-48 h-48 border border-gray-300">
            3
          </div>
        </CarouselItem>
      </CarouselContent>
      <CarouselPrevious />
      <CarouselNext />
    </Carousel>
  );
};
```

### DIFF
--- a/apps/docs.blocksense.network/components/common/Carousel.tsx
+++ b/apps/docs.blocksense.network/components/common/Carousel.tsx
@@ -1,0 +1,244 @@
+'use client';
+
+import {
+  useState,
+  useEffect,
+  useRef,
+  HTMLAttributes,
+  RefObject,
+  ReactNode,
+  createContext,
+  useContext,
+} from 'react';
+
+import { cn } from '@blocksense/ui/utils';
+import { Button } from '@blocksense/ui/Button';
+import { ImageWrapper } from '@blocksense/ui/ImageWrapper';
+import { ButtonProps } from '@blocksense/ui/Button';
+
+type CarouselProps = {
+  orientation?: 'horizontal' | 'vertical';
+  className?: string;
+  children?: ReactNode;
+};
+
+type CarouselContextProps = {
+  contentRef: RefObject<HTMLDivElement>;
+  scrollPrev: () => void;
+  scrollNext: () => void;
+  canScrollPrev: boolean;
+  canScrollNext: boolean;
+  orientation: 'horizontal' | 'vertical';
+};
+
+const CarouselContext = createContext<CarouselContextProps>(
+  {} as CarouselContextProps,
+);
+
+export const Carousel = ({
+  orientation = 'horizontal',
+  className,
+  children,
+  ...props
+}: CarouselProps) => {
+  const contentRef = useRef<HTMLDivElement>(null);
+  const [canScrollPrev, setCanScrollPrev] = useState(false);
+  const [canScrollNext, setCanScrollNext] = useState(false);
+  const [itemsCount, setItemsCount] = useState(1);
+  const [currentItem, setCurrentItem] = useState(1);
+
+  const updateScrollState = (currentIndex: number, totalItems: number) => {
+    setCanScrollNext(currentIndex < totalItems);
+    setCanScrollPrev(currentIndex > 1);
+  };
+
+  const scrollByAmount = (amount: number) => {
+    const container = contentRef.current;
+    if (container) {
+      if (orientation === 'horizontal') {
+        container.scrollBy({ left: amount, behavior: 'smooth' });
+      } else {
+        container.scrollBy({ top: amount, behavior: 'smooth' });
+      }
+    }
+  };
+
+  const scrollPrev = () => {
+    const container = contentRef.current;
+    if (container) {
+      const scrollAmount =
+        orientation === 'horizontal'
+          ? -container.clientWidth
+          : -container.clientHeight;
+      scrollByAmount(scrollAmount);
+
+      setCurrentItem(prevItem => {
+        const newItem = Math.max(prevItem - 1, 1);
+        updateScrollState(newItem, itemsCount);
+        return newItem;
+      });
+    }
+  };
+
+  const scrollNext = () => {
+    const container = contentRef.current;
+    if (container) {
+      const scrollAmount =
+        orientation === 'horizontal'
+          ? container.clientWidth
+          : container.clientHeight;
+      scrollByAmount(scrollAmount);
+
+      setCurrentItem(prevItem => {
+        const newItem = Math.min(prevItem + 1, itemsCount);
+        updateScrollState(newItem, itemsCount);
+        return newItem;
+      });
+    }
+  };
+
+  useEffect(() => {
+    const container = contentRef.current;
+    if (container) {
+      const currentItems = container.children.length;
+      setItemsCount(currentItems);
+      updateScrollState(currentItem, currentItems);
+    }
+  }, []);
+
+  return (
+    <CarouselContext.Provider
+      value={{
+        contentRef,
+        scrollPrev,
+        scrollNext,
+        canScrollPrev,
+        canScrollNext,
+        orientation,
+      }}
+    >
+      <div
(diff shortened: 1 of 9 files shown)


EmilIvanichkovv (commented):
Since this is a new component, there is absolutely no point in implementing it here and then moving it to UI lib. It seems like a waste of time. 

You can move this one in this PR, but for all following components, lets just implement it directly in the UI lib

VanyaPavlova:
> Since this is a new component, there is absolutely no point in implementing it here and then moving it to UI lib. It seems like a waste of time.
> 
> You can move this one in this PR, but for all following components, lets just implement it directly in the UI lib

Thank you for your consideration. This is a new component, part of the shared component lib, but it's part of another initiaive and it's separated differently into the project management tool, we're using. 

It makes sense, because the purpose of this task is for each developer to concentrate on the functionality of the component, following a common team agreement. Creating stories for a new component it's part of another initiative into the same PM tool. We're separating concerns here. So, for the second part, each developer is concentrating only on the stories for the component itself, without affecting the functionality, because it's already considered and approved at this stage.

Developing new component, creating stories in the same time is something we want to avoid, because it has bigger scope and it's easier to make mistakes.

This approach works well for the team so far, so are there any areas you'd like to improve?

EmilIvanichkovv:
> This is a new component, part of the shared component lib, but it's part of another initiative and it's separated differently into the project management tool, we're using.

To me, this highlights an issue in our project management approach. We're duplicating work, even though the end result remains the same.

> It makes sense, because the purpose of this task is for each developer to concentrate on the functionality of the component,

I understand that, and I agree with the idea of developers focusing on the component’s functionality. However, this specific component isn’t even used in the docs app. We could achieve the same level of focus by developing it directly in the UI library. In fact, testing and showcasing it in Storybook would give us a clearer view of its behavior and help us ensure it's implemented correctly.

> So, for the second part, each developer is concentrating only on the stories for the component itself, without affecting the functionality, because it's already considered and approved at this stage.

Again, this feels like a flawed approach. The development of a component and the exploration of its usage scenarios should go hand in hand—they're deeply connected and shouldn’t be separated.

> Developing new component, creating stories in the same time is something we want to avoid, because it has bigger scope and it's easier to make mistakes.

But in this case, we're developing something that doesn't get used at all. How can we be confident the component is actually correct? Also, migrating it later introduces room for error.

> This approach works well for the team so far, so are there any areas you'd like to improve?

As I mentioned before, I think we should develop the component directly in the UI library, write stories, and validate everything thoroughly. Once we know it's solid, we can then integrate it where it's needed.

VanyaPavlova:
> To me, this highlights an issue in our project management approach. We're duplicating work, even though the end result remains the same.

I personally tend to disagree. There is no duplicated work here, it's a separation of concerns, as I described above. 
Once we move forward to the design phase of the components, UX/UI guys, will be reviewing primary the stories initiative and it will be easier to orientate into the entire work that have to do for the stories.   

> I understand that, and I agree with the idea of developers focusing on the component’s functionality. However, this specific component isn’t even used in the docs app. We could achieve the same level of focus by developing it directly in the UI library. In fact, testing and showcasing it in Storybook would give us a clearer view of its behavior and help us ensure it's implemented correctly.

In fact, the initiative for the Storybook is clearly defined for components:
- used in the docs.
- not used in the docs.
- those who will be used across all projects (including upcoming).

The purpose of defining separate initiatives to have all components in the same place. You may go and see how tidy the items into the project are. 

> Again, this feels like a flawed approach. The development of a component and the exploration of its usage scenarios should go hand in hand—they're deeply connected and shouldn’t be separated.

See previous point :)))

> But in this case, we're developing something that doesn't get used at all. How can we be confident the component is actually correct? Also, migrating it later introduces room for error.

Actually, we have upcoming projects that will benefit the components, in general. Some projects will utilize them more, others won't. 

> As I mentioned before, I think we should develop the component directly in the UI library, write stories, and validate everything thoroughly. Once we know it's solid, we can then integrate it where it's needed.

Without having all tasks into centralized place into the PM tool, we risk to lose imporant information for this subproject itself.

Thank you.

RadoslavDimchev (approved):

VanyaPavlova (approved):

PR #1193, opened 2025-04-09
fix(nft.blocksense.network): Update Geist and GeistMono font configuration 

VanyaPavlova:
This PR contains a quick, fix that refines how Geist and Geist Mono fonts are loaded using next/font/local, without the need of using weight and evaluate it directly from CSS variables, in the way they're defined.


RadoslavDimchev (approved):
`font-weight` is working!

PR #1197, opened 2025-04-09
[nft]: Add `blocksense-favicon-dark` & use in metadata

RadoslavDimchev:
Blocksense dark favicon
image


VanyaPavlova:
We need to verify with marketing folks, the correct metadata description. Let's leave it in this way for now. Thank you! ❤️

PR #1200, opened 2025-04-09
fix(nft.blocksense.network): Add error.tsx and not-found.tsx pages

VanyaPavlova:


RadoslavDimchev (approved):
> We may update the style after the design review and make sure the layout is aligned correctly.

PR #1202, opened 2025-04-10
[docs.ws/ui]: Create `Carousel stories` to our `Storybook`

kalinsky-dev:
This PR aims to add `Carousel stories` to our `Storybook`.
The `chevron-left.svg` icon is also deleted, as it is only used in the `ui library` and not in the `docs.ws` workspace.


EmilIvanichkovv (commented):
You can add an example with images. Or any different items wrapped in the carousel component for better showcases

PR #1216, opened 2025-04-15
[nft]: Implement Form

RadoslavDimchev:
Implement Form section - Mint and Success.
- reusable components;
- responsive - desktop & mobile;
- components with states, status, and functions;

image
image


apps/nft.blocksense.network/app/design-system/autofill.css
```diff
@@ -0,0 +1,15 @@
...
+  textarea:-webkit-autofill:hover,
+  textarea:-webkit-autofill:focus,
+  select:-webkit-autofill,
+  select:-webkit-autofill:hover,
+  select:-webkit-autofill:focus {
+    -webkit-text-fill-color: var(--white);
+    transition: background-color 5000s ease-in-out 0s;
+    caret-color: var(--white);
+  }
+}
```
VanyaPavlova, line 15: Pretty decent work, even at this WIP phase. I'm leaving this as a comment here, because it might be interesting for someone.

It's better to move these from @layer utilities to @layer base.
Autofill pseudo-classes like :-webkit-autofill apply directly to native form elements and are not meant to be used as utility classes.

- @layer utilities is meant for reusable class-based styles.
- @layer base is the correct layer for element-level or browser-specific defaults like these.

You may leave the selectors as-is for clarity and compatibility, as presented below:
```

@layer base {
  input:-webkit-autofill,
  input:-webkit-autofill:hover,
  input:-webkit-autofill:focus,
  textarea:-webkit-autofill,
  textarea:-webkit-autofill:hover,
  textarea:-webkit-autofill:focus,
  select:-webkit-autofill,
  select:-webkit-autofill:hover,
  select:-webkit-autofill:focus {
    -webkit-text-fill-color: var(--white);
    transition: background-color 5000s ease-in-out 0s;
    caret-color: var(--white);
  }
}

```

There is also possibility to use shorthand and to avoid repetion, as presented below.
```
@layer base {
  :is(input, textarea, select):-webkit-autofill:is(:hover, :focus, :not(:hover):not(:focus)) {
    -webkit-text-fill-color: var(--white);
    transition: background-color 5000s ease-in-out 0s;
    caret-color: var(--white);
  }
}
```

Even with the first option is pretty fine. Thanks.

apps/nft.blocksense.network/components/Input.tsx
```diff
@@ -0,0 +1,87 @@
...
+      ? 'border-[var(--red)]'
+      : status === 'success'
+        ? 'border-[var(--green)]'
+        : 'border-[var(--gray-dark)]';
+
+  return (
+    <article className="input__container flex flex-col gap-2">
+      <label
+        htmlFor={id}
+        className={`input__label flex md:gap-3 gap-2 items-center rounded-2xl md:px-4 md:py-5 p-4 text-[var(--gray-medium)] border ${borderColor} hover:border-[var(--gray-light)] focus-within:border-[var(--gray-light)] ${labelClassName}`}
```
VanyaPavlova, line 53: Add transition for the states, as presented below. Also, it would be great if hover and focus are bit more distinguishable, so feel free to ask the designer for that, after adding the transition. 
```suggestion
 className={`input__label flex md:gap-3 gap-2 items-center rounded-2xl md:px-4 md:py-5 p-4 text-[var(--gray-medium)] border ${borderColor} hover:border-[var(--gray-light)] focus-within:border-[var(--gray-light)] transition duration-300 ease-in-out ${labelClassName}`}
```

apps/nft.blocksense.network/app/design-system/autofill.css
```diff
@@ -0,0 +1,15 @@
...
+  input:-webkit-autofill:hover,
+  input:-webkit-autofill:focus,
+  textarea:-webkit-autofill,
+  textarea:-webkit-autofill:hover,
+  textarea:-webkit-autofill:focus,
+  select:-webkit-autofill,
+  select:-webkit-autofill:hover,
+  select:-webkit-autofill:focus {
+    -webkit-text-fill-color: var(--white);
+    transition: background-color 5000s ease-in-out 0s;
```
VanyaPavlova, line 12: Probably it's a typo, on this line, but let's mention it. A duration of 5000 seconds is a bit long. 
Transitions are supposed to make things look smooth and snappy. 

In most cases, we use predefined timing from 100ms to 500ms for quick transitions that catch the eye.

RadoslavDimchev: More info [here](https://stackoverflow.com/questions/71640408/trouble-with-input-styling-and-autofill) why I chose 5000s

VanyaPavlova: Let's leave it like that, but remember that it's not something we use on everyday basis.

VanyaPavlova (approved):

PR #1219, opened 2025-04-15
build(typescript): Upgrade toolchain and fix and improve rollup builder

PetarKirov:
## Toolchain upgrades

* Node.js: `22.14.0` -> `23.11.0`
* Yarn: `4.5.0` -> `4.9.1`
* TypeScript: `5.5.4` -> `5.8.3`
* Storybook: `8.5.6` -> `8.6.12`

---

#### Details

* To upgrade TypeScript we had to upgrade Yarn, as it includes patches to the TypeScript compiler, like this one: https://github.com/yarnpkg/berry/pull/6591
* To upgrade Yarn we had to upgrade Node.js to >= v23.5.0, which is the first version that supports [`module.registerHooks()`](https://nodejs.org/api/module.html#moduleregisterhooksoptions)
  * For details read:
    * [An analysis from Yarn maintainer](https://github.com/yarnpkg/berry/issues/6646#issuecomment-2656674001) on the complicated state of affairs of the Node.js module customization hooks
    * [Node.js's `module.registerHooks()` tracking issue](https://github.com/nodejs/node/issues/56241)

* Switch storybook framework from `@storybook/nextjs` to `@storybook/experimental-nextjs-vite` (from Webpack to Vite)
  * This was necessary as the above mentioned toolchain upgrades cause module resolution errors

## `apps/rollup` builder improvements

* Enable source map rollup output based on tsconfig
  * This removes the following warning: image
* Remove "Generated an empty chunk" warning which occur when transpiling a module containing only TS types
  * Example: image
* Ensure the build fails on typechecking errors
  * Previously, an TS error in base-utils could go completely unnoticed: image

### CLI UX

| Command | Before | After |
|--------|--------|--------|
| `yarn build-single @blocksense/base-utils` | image | image |
| `yarn build-single @blocksense/base-utils` error | image | image |
| `yarn build @blocksense/data-feeds-config-generator` error | image | image |
| `yarn build @blocksense/data-feeds-config-generator` | image | image |
| `just build-ts` | image | image |


nix/shells/pkg-sets/js.nix
```diff
@@ -1,17 +1,13 @@
...
-    yarn = oldYarn;
-  };
+  corepack = pkgs.corepack.override { inherit nodejs; };
 in
 {
   packages =
     [
       nodejs
-      yarn
+      corepack
```
EmilIvanichkovv, line 10: cool

apps/rollup/bundle.mjs
```diff
@@ -109,6 +109,7 @@ function createConfig(packageDir, format) {
         dir: `${packageDir}/dist/${format}`,
         format: format,
         entryFileNames: `[name].${{ esm: 'mjs', cjs: 'cjs' }[format]}`,
+        sourcemap: tsconfig.config.compilerOptions.sourceMap,
```
EmilIvanichkovv, line 138: Why we need this change?

Justfile
```diff
@@ -3,11 +3,18 @@ root-dir := justfile_directory()
 default:
   @just --list
 
-build-ts:
-  yarn build-single @blocksense/base-utils
-  yarn build-single @blocksense/config-types
-  yarn build-single @blocksense/sol-reflector
-  yarn build-single @blocksense/data-feeds-config-generator
+build-ts package="all":
```
EmilIvanichkovv, line 6: nice improvement!

Justfile
```diff
@@ -3,11 +3,18 @@ root-dir := justfile_directory()
...
+  #!/usr/bin/env bash
+  set -euo pipefail
+  if [[ {{package}} == "all" ]]; then
+    yarn build-single @blocksense/base-utils
+    yarn build-single @blocksense/config-types
+    yarn build-single @blocksense/sol-reflector
+    yarn build-single @blocksense/contracts
```
EmilIvanichkovv, line 13: I have some concerns regarding `@blocksense/contracts`.

On one hand, we already have a dedicated CI job for smart contracts—similar to what we have for the websites—which is why I initially excluded them from this workflow.  

On the other hand, this package contains a significant amount of TypeScript code that still needs to be type-checked, built, and properly maintained. That’s why I’m wondering whether we should include it here as well.

apps/rollup/bundle.mjs
```diff
@@ -24,36 +25,47 @@ function main() {
...
     }
   } catch (error) {
-    console.error(error);
+    console.log(
+      error.message
+        .replace(
+          /^\[plugin typescript\]\s+(\S+)\s+\((\d+:\d+)\):\s+@rollup\/plugin-typescript\s+(TS\d+):/,
+          '├─ ❌ $1:$2 - error $3:',
+        )
+        .replace(/^\s+/gm, '│       '),
```
EmilIvanichkovv, line 65: uff :sob:

apps/rollup/bundle.mjs
```diff
@@ -24,36 +25,47 @@ function main() {
...
-      fs.rename(`${packageDir}/dist/esm/types`, `${packageDir}/dist/types`),
-    )
-    .then(() => console.log('Build finished successfully'));
+  await Promise.all([
+    build(packageDir, relativeDir, 'esm'), //
+    build(packageDir, relativeDir, 'cjs'),
+  ]);
+  await fs.rename(`${packageDir}/dist/esm/types`, `${packageDir}/dist/types`);
+  console.log(
+    chalkTemplate`╰─ Finished {bold ${relativeDir}/dist/\{cjs,esm,types\}} successfully.`,
```
EmilIvanichkovv, line 45: nice

nix/filesets.nix
```diff
@@ -17,11 +17,16 @@ with lib.fileset;
...
         ]
       ) root)
 
+      # JSON files must be listed one by one, otherwise changing an
+      # unrelated JSON file will cause all Rust derivations to be rebuilt
+      (root + "/apps/sequencer_tests/Safe.json")
+      (root + "/apps/oracles/eth-rpc/src/abi/YnETHx.json")
+      (root + "/apps/sequencer_tests/SafeProxyFactory.json")
+      (root + "/libs/gnosis_safe/safe_abi.json")
+      (root + "/apps/oracles/eth-rpc/src/abi/YnETHx.json")
```
EmilIvanichkovv, line 29: This will reduce the build time, right?

EmilIvanichkovv (commented):
LGTM

Have some questions

EmilIvanichkovv:
The CI fails tho.

nix/filesets.nix
Xearty: It only affects when the rust project gets rebuilt, not how long it builds

nix/shells/pkg-sets/js.nix
PetarKirov: `corepack` is in charge for downloading Node.js package managers so we no longer need to download `yarn` via Nix. This way we can more easily upgrade, simply by changing the `packageManager` field in the root `package.json` file.

Justfile
PetarKirov: Note that with this change we're still not type checking the TS code in the contracts. This is something I'll do I. the next PR

apps/rollup/bundle.mjs
PetarKirov: Because of this: https://github.com/rollup/plugins/blob/typescript-v12.1.2/packages/typescript/src/options/validate.ts#L22

as mentioned above:

> * This removes the following warning: image

Xearty (approved):

Justfile
PetarKirov: This change is aimed more for local development, than CI. In the CI we're duplicating work a few times already, as the docs.ws builds the contracts, for example.
Running `yarn build @blocksense/data-feeds-config-generator` doesn't build the contracts, which may give a false sense of security when making changes to `@blocksense/config-types`.

EmilIvanichkovv (approved):
Seems fine.
image


`yarn build:all` still not works but this is not in the scope of this PR

VanyaPavlova (approved):

PR #1225, opened 2025-04-15
[nft]: Update colors

RadoslavDimchev:
- #2B2929 -> #262626
- Add default `white` color for H titles and input

### DIFF
--- a/apps/nft.blocksense.network/app/design-system/colors.css
+++ b/apps/nft.blocksense.network/app/design-system/colors.css
@@ -2,7 +2,7 @@
   Define global color tokens
 */
 :root {
-  --black: #2b2929;
+  --black: #262626;
   --white: #fffbfa;
   --neon-yellow: #eeff00;
   --gray-light: #d9d9d9;
--- a/apps/nft.blocksense.network/app/design-system/colors.css
+++ b/apps/nft.blocksense.network/app/design-system/colors.css
@@ -13,6 +13,5 @@
 @layer base {
   body {
     background: var(--black);
-    color: var(--gray-light);
   }
 }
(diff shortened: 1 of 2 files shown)


apps/nft.blocksense.network/app/design-system/colors.css
```diff
@@ -15,4 +15,11 @@
     background: var(--black);
     color: var(--gray-light);
   }
+
+  h1,
+  h2,
+  h3,
+  h4 {
+    color: var(--white);
+  }
```
VanyaPavlova, line 24: Since, it's a bit late, I'm leaving that here - let's move h1, h2, h3, h4 updates to [/design-system/typography/default.css](https://github.com/blocksense-network/blocksense/blob/main/apps/nft.blocksense.network/app/design-system/typography/default.css).

RadoslavDimchev: Do you want to move `color: var(--gray-light)` for the `body` typography?

Also, the input color should be moved.

VanyaPavlova: @RadoslavDimchev yes. The [colors.css](https://github.com/blocksense-network/blocksense/blob/main/apps/nft.blocksense.network/app/design-system/colors.css) is primarily meant to be holding colors only. Especially for the typography, basic settings into the typography file. All other custom styles, that might appear and don't relate to typography, should be moved to [globalls.css](https://github.com/blocksense-network/blocksense/blob/main/apps/nft.blocksense.network/app/globals.css). Thank you.

VanyaPavlova (approved):

kalinsky-dev (approved):

PR #1250, opened 2025-04-22
fix(sequencer_tests): In tests, we need to configure sequencer not to use gnosis safe at this point

HristoStaykov:


melatron (approved):
LGTM

PR #1253, opened 2025-04-23
[nft]: Social verification

RadoslavDimchev:


apps/nft.blocksense.network/workers/socialDataAPI.ts
```diff
@@ -0,0 +1,60 @@
+'use server';
+
+const BLOCKSENSE_ID = process.env['NEXT_PUBLIC_BLOCKSENSE_X_ID'];
+const API_KEY = process.env['NEXT_PUBLIC_SOCIAL_DATA_API_KEY'];
+const TWEET_ID = '1913586022836785624'; // TODO: with the actual tweet ID
+
+export async function getUserIdByUsername(username: string) {
```
EmilIvanichkovv, line 7: While testing, I used `@E_Ivanichkov` as my X handle — I had just opened my account and copied what I assumed was my handle (since the form asked for "X handle"). But I got a `You are not following us on X` error. Once I removed the `@`, everything worked.

So I’d suggest:
1. Improving the UI/UX a bit here:
   image
2. Stripping the `@` automatically in the code.

EmilIvanichkovv, line 7:
```suggestion
export async function getUserIdByXHandle(xHandle: string) {
```

apps/nft.blocksense.network/workers/socialDataAPI.ts
```diff
@@ -0,0 +1,60 @@
...
+  const response = await fetch(
```
EmilIvanichkovv, line 8:
```suggestion
  // Remove the '@' from the xHandle
  const username = xHandle.slice(1);
  const response = await fetch(
```

EmilIvanichkovv: With that said we can update `onXHandleChange` as follows:
```ts
  const onXHandleChange = (e: ChangeEvent<HTMLInputElement>) => {
    const value = e.target.value;
    setXHandle(value);

    // Regex to check if the input starts with '@'
    if (!/^@/.test(value)) {
        setXError('Incorrect X handle. It should start with "@"');
        setXSuccess('');
    } else {
        setXError('');
    }
};
```

EmilIvanichkovv (changes requested):

apps/nft.blocksense.network/workers/socialDataAPI.ts
EmilIvanichkovv: After an offline discussion, we decided it's best to allow users to set both `@hanele` and `handle` here

apps/nft.blocksense.network/components/MintForm.tsx
EmilIvanichkovv, line 1: I find the placeholder of the Discord input box misleading. Consider removing the `#` and using something like `Your Discord username`

EmilIvanichkovv (approved):

EmilIvanichkovv (approved):

EmilIvanichkovv (approved):

PR #1257, opened 2025-04-24
[oracles]: Implement initial version of `stock-price-feeds` oracle

EmilIvanichkovv:
This PR introduces the first functional version of the `stock-price-feeds` oracle.

---

### Core Features

- Implements the main oracle pipeline:
  - Fetches data from multiple providers concurrently.
  - Computes VWAP from provider-level price points.
  - Outputs payloads in the standard `blocksense-sdk` format.

- Adds support for two providers:
  - **AlphaVantage** ( end-of-day data, little daily requests )
  - **YahooFinance** ( real-time data )

Both fetchers read their respective API keys from the `capabilities` section of `Settings`.

---

### SDK & Utility Enhancements

- Generalizes the `PricesFetcher` trait to accept an optional `api_key`.
- Introduces a helper `get_api_key(...)` to extract provider credentials.
- Adds support for optional HTTP headers in `http_get_json(...)`.

---

### Configuration and Setup

- Adds initial feed configuration to `spin.toml` and `feeds_config_v2.json`, including `providers` under `arguments`.
- Declares outbound hosts for both providers.
- Introduces the `Capabilities` type and propagates it through the fetch pipeline.
- Adds dependencies: `futures` and `itertools`.

---

This PR sets the foundation for extending `stock-price-feeds` with additional providers in future iterations.

### DIFF
--- a/apps/oracles/crypto-price-feeds/src/exchanges/binance.rs
+++ b/apps/oracles/crypto-price-feeds/src/exchanges/binance.rs
@@ -43,6 +43,7 @@ impl<'a> PricesFetcher<'a> for BinancePriceFetcher<'a> {
             let response = http_get_json::<BinancePriceResponse>(
                 "https://api1.binance.com/api/v3/ticker/24hr",
                 Some(&[("symbols", req_symbols.as_str())]),
+                None,
             )
             .await?;
 
--- a/apps/oracles/crypto-price-feeds/src/exchanges/binance.rs
+++ b/apps/oracles/crypto-price-feeds/src/exchanges/binance.rs
@@ -29,7 +29,7 @@ pub struct BinancePriceFetcher<'a> {
 impl<'a> PricesFetcher<'a> for BinancePriceFetcher<'a> {
     const NAME: &'static str = "Binance";
 
-    fn new(symbols: &'a [String]) -> Self {
+    fn new(symbols: &'a [String], _api_key: Option<&'a str>) -> Self {
         Self { symbols }
     }
 
--- a/apps/oracles/stock-price-feeds/src/fetch_prices.rs
+++ b/apps/oracles/stock-price-feeds/src/fetch_prices.rs
@@ -8,9 +8,9 @@ use futures::stream::{FuturesUnordered, StreamExt};
 use blocksense_sdk::traits::prices_fetcher::{fetch, TradingPairSymbol};
 
 use crate::{
-    providers::alpha_vantage::AlphaVantagePriceFetcher,
+    providers::{alpha_vantage::AlphaVantagePriceFetcher, yahoo_finance::YFPriceFetcher},
     types::{
-        Capabilities, ProviderPriceData, ProvidersSymbols, PairToResults, ResourceData,
+        Capabilities, PairToResults, ProviderPriceData, ProvidersSymbols, ResourceData,
         ResourcePairData,
     },
     utils::get_api_key,
@@ -20,6 +20,7 @@ use crate::{
 #[serde(deny_unknown_fields)]
 pub struct SymbolsData {
     pub alpha_vantage: Vec<TradingPairSymbol>,
+    pub yahoo_finance: Vec<TradingPairSymbol>,
 }
 
 impl SymbolsData {
@@ -27,6 +28,10 @@ impl SymbolsData {
         Ok(Self {
             alpha_vantage: providers_symbols
                 .get("AlphaVantage")
+                .cloned()
+                .unwrap_or_default(),
+            yahoo_finance: providers_symbols
+                .get("YahooFinance")
                 .cloned()
                 .unwrap_or_default(),
         })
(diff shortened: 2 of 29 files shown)


apps/oracles/stock-price-feeds/src/fetch_prices.rs
```diff
@@ -0,0 +1,97 @@
...
+            Ok(prices) => {
+                let time_taken = before_fetch.elapsed();
+                println!("ℹ️  Successfully fetched prices from {market_id} in {time_taken:?}",);
+                let prices_per_market = MarketPriceData {
+                    name: market_id.to_owned(),
+                    data: prices,
+                };
+                println!("Prices from {market_id}: {prices_per_market:?}");
+                fill_results(&resources.pairs, prices_per_market, &mut results);
+                println!("Results: {results:?}");
```
EmilIvanichkovv, line 66: remove

apps/oracles/stock-price-feeds/src/fetch_prices.rs
```diff
@@ -0,0 +1,97 @@
...
+    while let Some((market_id, result)) = futures_set.next().await {
+        match result {
+            Ok(prices) => {
+                let time_taken = before_fetch.elapsed();
+                println!("ℹ️  Successfully fetched prices from {market_id} in {time_taken:?}",);
+                let prices_per_market = MarketPriceData {
+                    name: market_id.to_owned(),
+                    data: prices,
+                };
+                println!("Prices from {market_id}: {prices_per_market:?}");
```
EmilIvanichkovv, line 64: remove

yordanmadzhunkov (approved):

apps/oracles/crypto-price-feeds/src/exchanges/binance.rs
```diff
@@ -29,7 +29,7 @@ pub struct BinancePriceFetcher<'a> {
 impl<'a> PricesFetcher<'a> for BinancePriceFetcher<'a> {
     const NAME: &'static str = "Binance";
 
-    fn new(symbols: &'a [String]) -> Self {
+    fn new(symbols: &'a [String], __api_key: Option<&'a str>) -> Self {
```
melatron, line 32: NIT: double underscore looks a bit strange, I prefer using only one.

libs/sdk/src/traits/prices_fetcher.rs
melatron, line 1: NIT: Not a huge fan of this pattern, but for now it works. Overall, we need to have more general SDK helper traits and functions which directly return the WIT types that the oracle script guest function implementation returns. We need to organize a design meeting in which we should shake hands on the perfect flow and then implement it accordingly. Initial implementation of this was here - https://github.com/blocksense-network/blocksense/commit/e4edb8c48b033765a885500f7a8d5aceafc3f399 (price_pair.rs). It was a starting point. 
We also need to handle different types of result aggregation or fallback and streamline the whole flow of writing a price feed oracle. Lets discuss this further in a meeting.

apps/oracles/stock-price-feeds/src/providers/yahoo_finance.rs
```diff
@@ -0,0 +1,100 @@
...
+                                if price.is_none() { "price" } else { "" },
+                                if price.is_none() && volume.is_none() {
+                                    " and "
+                                } else {
+                                    ""
+                                },
+                                if volume.is_none() { "volume" } else { "" },
+                            );
+                            None
+                        }
```
melatron, line 93: NIT: We have the option to return Error type result in the oracle scripts. Is there a reason we are not doing it anymore? Please also lets create a variable from the result of this `filter_map().collect()` and then return it as `Ok(variable_with_proper_name)`.

apps/oracles/stock-price-feeds/src/lib.rs
```diff
@@ -1,12 +1,216 @@
...
+    println!("[{}]", missing_prices);
+
+    println!("\nResults:");
+    table.printstd();
+}
+
+/*TODO:(EmilIvanichkovv):
+    This is a copy-paste from the `crypto-price-feeds` oracle.
+    It should be moved to blocksense-sdk
+*/
```
melatron, line 207: <3

melatron (approved):
With some NITs that can be ignored.

libs/sdk/src/traits/prices_fetcher.rs
EmilIvanichkovv: Agree

apps/oracles/stock-price-feeds/src/providers/yahoo_finance.rs
EmilIvanichkovv: > Please also lets create a variable from the result of this filter_map().collect() and then return it as Ok(variable_with_proper_name).

Will do

EmilIvanichkovv: > We have the option to return an `Error` type result in the oracle scripts. Is there a reason we're not doing that here?

The data fetched here is **not the final output** of the oracle script. This part is specifically for fetching data from Yahoo Finance—just one of potentially many providers we'll support in the future. The values retrieved here go through additional processing before the script decides whether to return an `Error`.

At this stage, I just want to **log any missing data** from this provider for further investigation.

melatron (approved):
LGTM, well done.

PR #1259, opened 2025-04-24
refactor(feeds/types/enconding): Refactor encoding of FeedType to return Result

HristoStaykov:


melatron (approved):
LGTM
NIT:
* `thiserror` crate is ideal for library development where detailed information is beneficial for users (programmers).
* `anyhow` crate is suited for applications where internal details are not crucial, providing simplified information to users.

PR #1263, opened 2025-04-25
[nft/refactor]: Organize `Intersection Observer functionality` to prevent `Code Duplication`

kalinsky-dev:
### Summary

This PR introduces a reusable hook, `useRevealOnView`, to centralize and simplify the logic for triggering animations and bringing elements into view using the `IntersectionObserver` API. Previously, this logic was duplicated across multiple components, including the Hero Section, Form Block, About NFT Section, CTA Section, and Footer. By consolidating this functionality into a single hook, we improve code consistency, maintainability, and readability.

### Changes

1. **Added**: `useRevealOnView.ts` in `/hooks` to encapsulate the `IntersectionObserver` logic.
2. **Updated**: Components (`Hero.tsx`, `Form.tsx`, `About.tsx`, `CTA.tsx`, `Footer.tsx`) to use the new `useRevealOnView` hook.
3. **Removed**: Inline `IntersectionObserver` logic from the above components to eliminate code duplication.

### `useRevealOnView` Hook

---

#### **Parameters**

- **`threshold`** (optional):
  - **Type**: `number`
  - **Default**: `0.1`
  - **Description**: A number between `0` and `1` that specifies how much of the element must be visible in the viewport to trigger the `show` state. For example:
    - `0.1` means 10% of the element must be visible.
    - `1` means the entire element must be visible.
---
#### **Return Values**

The hook returns an object with the following properties:

- **`targetRef`**:
  - **Type**: `RefObject<HTMLDivElement>`
  - **Description**: A `ref` to attach to the element you want to observe.

- **`show`**:
  - **Type**: `boolean`
  - **Description**: A boolean indicating whether the element is currently visible in the viewport.
---

#### **Usage Example**
```tsx
'use client';

import { useRevealOnView } from 'hooks/useRevealOnView';

export const ExampleComponent = () => {
  const { targetRef, show } = useRevealOnView(0.5); // Trigger when 50% of the element is visible

  return (
    <div
      ref={targetRef}
      className={`transition-opacity duration-500 ${
        show ? 'opacity-100' : 'opacity-0'
      }`}
    >
      This content fades in when it comes into view.
    </div>
  );
};

### DIFF
--- a/apps/nft.blocksense.network/components/About.tsx
+++ b/apps/nft.blocksense.network/components/About.tsx
@@ -1,6 +1,5 @@
 'use client';
 
-import { useEffect, useRef, useState } from 'react';
 import Image from 'next/image';
 
 import aboutVector from '/public/images/about-vector.svg';
@@ -13,6 +12,8 @@ import bsxParrotMobile from '/public/images/bsx-parrot-mobile.png';
 import bsxRobotMobile from '/public/images/bsx-robot-mobile.png';
 import bsxWarriorMobile from '/public/images/bsx-warrior-mobile.png';
 
+import { useRevealOnView } from 'hooks/useRevealOnView';
+
 export const About = () => {
   return (
     <>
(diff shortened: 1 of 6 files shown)


RadoslavDimchev:
Good job, it's clearer, but in reference to #1277 it needs to be closed

PR #1276, opened 2025-04-29
[oracles/stock price feeds]: Implement fetcher for `Twelvedata`

MillaGenova:


apps/oracles/stock-price-feeds/src/providers/twelvedata.rs
```diff
@@ -0,0 +1,86 @@
...
+                            eprintln!(
+                                "[TwelveData] Skipping symbol {}: missing {}{}{}",
+                                value.symbol,
+                                if price.is_none() { "price" } else { "" },
+                                if price.is_none() && volume.is_none() {
+                                    " and "
+                                } else {
+                                    ""
+                                },
+                                if volume.is_none() { "volume" } else { "" },
```
EmilIvanichkovv, line 75: nit: It seems we use this log in most of the providers' fetchers. In the future, we can export it to `apps/oracles/stock-price-feeds/src/utils.rs`

EmilIvanichkovv (approved):
While reviewing, I hit [[Twelve Data’s rate limit of 8 API calls per minute](https://twelvedata.com/pricing)].
Here’s the response they return when the limit is exceeded:

```json
{
  "code": 429,
  "message": "You have run out of API credits for the current minute. 12 API credits were used, with the current limit being 8. Wait for the next minute or consider switching to a higher tier plan at https://twelvedata.com/pricing",
  "status": "error"
}
```

In production, we likely won’t hit this limit—but still, it’d be good to improve how we handle this case in the oracle.

Currently, we log:
```
❌ Error fetching prices from twelvedata: invalid type: integer `429`, expected struct PriceData at line 1 column 11
```

We could handle this more gracefully by improving the `http_get_json` function ([link](https://github.com/blocksense-network/blocksense/blob/main/libs/sdk/src/http.rs#L36)).  
Let’s leave this for now but keep it in mind as a follow-up improvement.

PR #1279, opened 2025-04-30
Add NixOS VM integration test to the Blocksense CI (WIP)

monyarm:
- **refactor(nix/test-environments): Remove process-compose (devenv) from example-setup-01 and example-setup-02**
- **feat(test-environments): Move process-compose related options to a seperate module.**
- **feat(test-environments): Add basic NixOS Integration test, which tests if the services in example-setup-01 start.**


.github/workflows/ci.yml
```diff
@@ -273,3 +273,27 @@ jobs:
...
+
+      - uses: cachix/cachix-action@v16
+        with:
+          name: ${{ vars.CACHIX_CACHE }}
+          authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+
+      - name: Build & activate the Nix Dev Shell
+        run: |
+          eval "$(nix print-dev-env --accept-flake-config --impure .#devShells.x86_64-linux.js || echo exit 1)"
+          env >> "$GITHUB_ENV"
```
PetarKirov, line 296: See if this CI step can be skipped. I would expect that we don't need to the tools provided by the JavaScript dev shell for the next one.

.github/workflows/ci.yml
```diff
@@ -273,3 +273,27 @@ jobs:
...
+
+      - name: Run NixOS tests
+        run: nix eval --json .#legacyPackages.x86_64-linux.nixosTests --apply "builtins.attrNames" | jq -r '.[]' | xargs -I {} nix build --option sandbox false -L --json --accept-flake-config .#legacyPackages.x86_64-linux.nixosTests.{}
```
PetarKirov, line 299: Can we use `mcl ci_matrix` for this? Also why do we need to disable the sandbox?

.github/workflows/trigger-downstream-flake-update.yml
```diff
@@ -16,13 +16,38 @@ jobs:
         include:
           - downstream-repo: 'blocksense-network/infra'
     steps:
+      - name: Compare nix outputs
+        id: compare
+        run: |
+          # Define your command here (replace this example)
+          CMD="nix eval --raw .\#nixosConfigurations.example-setup-01.config.system.build.toplevel.outPath1"
```
PetarKirov, line 23:
```suggestion
          function get_machine_outpath {
            commit="$1"
            config="$2"
            nix eval --raw "github:blocksense-network/blocksense?rev=${commit}#nixosConfigurations.${config}.config.system.build.toplevel.outPath1"
          }
```

.github/workflows/trigger-downstream-flake-update.yml
```diff
@@ -16,13 +16,38 @@ jobs:
...
+
+          # Get current output
+          CURRENT_OUTPUT=$(eval "$CMD")
+          echo "Current output: $CURRENT_OUTPUT"
+
+          # Get previous commit output
+          git checkout ${{ github.event.before }}
+          PREVIOUS_OUTPUT=$(eval "$CMD")
+          echo "Previous output: $PREVIOUS_OUTPUT"
+          git checkout -
```
PetarKirov, line 33:
```suggestion
          PREVIOUS_OUTPUT="$(get_machine_outpath "${{ github.event.before }}" "example-setup-01")"
          echo "After: '$PREVIOUS_OUTPUT'"
```

.github/workflows/trigger-downstream-flake-update.yml
```diff
@@ -16,13 +16,38 @@ jobs:
...
     steps:
+      - name: Compare nix outputs
+        id: compare
+        run: |
+          # Define your command here (replace this example)
+          CMD="nix eval --raw .\#nixosConfigurations.example-setup-01.config.system.build.toplevel.outPath1"
+
+          # Get current output
+          CURRENT_OUTPUT=$(eval "$CMD")
+          echo "Current output: $CURRENT_OUTPUT"
```
PetarKirov, line 27:
```suggestion
          CURRENT_OUTPUT="$(get_machine_outpath "${{ github.sha }}" "example-setup-01")"
          echo "After: '$CURRENT_OUTPUT'"
```

PetarKirov (commented):
First review pass:

.github/workflows/trigger-downstream-flake-update.yml
PetarKirov: On second thought, can you implement my suggestion as a `just` recipe, so it could be used outside the CI?

nix/modules/blocksense/backends/process-compose.nix
```diff
@@ -44,7 +44,7 @@ let
           timeout_seconds = 30;
         };
         log_configuration = logsConfig;
-        log_location = cfg.logsDir + "/anvil-${name}.log";
+        log_location = if (builtins.isNull cfg.logsDir) then null else cfg.logsDir + "/anvil-${name}.log";
```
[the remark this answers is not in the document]
monyarm: It was complaining about logsDir being null, when I did a direnv reload. Which probably has to do with the test-environments being loaded in the shell, and the fact that logsDir was made null by default.

.github/workflows/ci.yml
zah: As a general rule, we should avoid long and complicated commands in the CI workflow files.
The command can be given a short descriptive name and added to the `Justfile`, so it's easy to reproduce the CI test locally as well.

Justfile
```diff
@@ -33,6 +33,11 @@ start-oracle oracle-name:
   cd "{{root-dir}}/apps/oracles/{{oracle-name}}"
   RUST_LOG=trigger=info "${SPIN:-spin}" build --up
 
+get-machine-outpath commit machine:
+  #!/usr/bin/env bash
+  set -euo pipefail
+  nix eval --raw "github:blocksense-network/blocksense?rev={{commit}}#nixosConfigurations.{{machine}}.config.system.build.toplevel.outPath"
```
PetarKirov, line 39: * consider renaming `get-machine-outpath` to `eval-machine`, as we will probably add `build-machine` and `start-machine` in the future
* Make the `commit` parameter last and make it optional
* Allow building with local (uncommitted) changes 


```suggestion
eval-machine machine commit="working-tree":
  #!/usr/bin/env bash
  set -euo pipefail
  if [[ commit = "working-tree" ]; then
    nix eval --raw "{{ root-dir }}#nixosConfigurations.{{machine}}.config.system.build.toplevel.outPath"
  else
    nix eval --raw "github:blocksense-network/blocksense?rev={{commit}}#nixosConfigurations.{{machine}}.config.system.build.toplevel.outPath"
  fi
```

.github/workflows/trigger-downstream-flake-update.yml
```diff
@@ -16,13 +16,33 @@ jobs:
         include:
           - downstream-repo: 'blocksense-network/infra'
     steps:
+      - name: Compare nix outputs
+        id: compare
+        run: |
+          # Get current output
+          CURRENT_OUTPUT="$(get_machine_outpath "${{ github.sha }}" "example-setup-01")"
```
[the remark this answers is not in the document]
monyarm: I missed the "just" on the CURRENT_OUTPUT command. I will resolve it.

nix/test-environments/example-setup-01.nix
```diff
@@ -1,15 +1,17 @@
...
 }:
 let
   # Function to read and parse the JSON file
   readJson = path: builtins.fromJSON (builtins.readFile path);
 
-  testKeysDir = config.devenv.root + "/nix/test-environments/test-keys";
-  deploymentV1FilePath = config.devenv.root + "/config/evm_contracts_deployment_v1.json";
-  deploymentV2FilePath = config.devenv.root + "/config/evm_contracts_deployment_v2/ink-sepolia.json";
+  testEnvironments = ./.;
+  configDir = ../../config;
```
EmilIvanichkovv, line 10: This might make conflicts with the work of @reo101 at https://github.com/blocksense-network/blocksense/pull/1478

nix/test-environments/example-setup-01.nix
```diff
@@ -19,14 +21,9 @@ let
   impersonationAddress = lib.strings.fileContents "${testKeysDir}/impersonation_address";
 in
 {
-  services.kafka = {
-    enable = true;
-  };
```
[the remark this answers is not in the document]
monyarm: It didn't exist when I initially made this PR. I'll revert it.

nix/test-environments/example-setup-01.nix
monyarm: Not sure how?

Justfile
```diff
@@ -128,6 +128,28 @@ start-oracle oracle-name:
...
+  nix eval --raw "${NIX_TARGET}#nixosConfigurations.{{machine}}.config.system.build.toplevel.outPath"
+
+[group('General')]
+[doc('Run NixOS integration tests')]
+run-nix-tests:
+  #!/usr/bin/env bash
+  set -euo pipefail
+  nix eval --json .#legacyPackages.x86_64-linux.nixosTests --apply "builtins.attrNames" | \
+    jq -r '.[]' | \
+    xargs -I {} \
```
PetarKirov, line 150: Extract this into a `list-nixos-tests` just recipe, along the lines of `list-environments`

Justfile
```diff
@@ -128,6 +128,28 @@ start-oracle oracle-name:
...
+    nix build --option sandbox false -L --json --accept-flake-config .#legacyPackages.x86_64-linux.nixosTests.{}
```
PetarKirov, line 151: Run this command in a for loop so that you can print the name test

nix/machines/default.nix
```diff
@@ -0,0 +1,68 @@
...
+  allEnvironmentPaths = lib.filter (path: lib.hasPrefix "example-setup-" (builtins.baseNameOf path)) (
+    lib.attrValues (builtins.readDir ../test-environments)
+  );
+
+  systems = [
+    "x86_64-linux"
+    "aarch64-linux"
+    "x86_64-darwin"
+    "aarch64-darwin"
+  ];
```
reo101, line 16: Are these all the systems we're going to test for? Can we source this from somewhere else, maybe sync it up with the global `flake-parts` `systems` list?

nix/machines/default.nix
```diff
@@ -0,0 +1,68 @@
...
+      baseName = builtins.baseNameOf path;
+      machinesForPath = lib.listToAttrs (
+        lib.map (system: {
+          name = if system == "x86_64-linux" then baseName else "${baseName}-${system}";
+          value = testEnvironmentConfig path system;
+        }) systems
+      );
+    in
+    acc // machinesForPath
+  ) { } allEnvironmentPaths;
```
reo101, line 29: Why the preference for `x86_64-linux`? Is it not better to make them the same?

nix/test-environments/default.nix
```diff
@@ -102,5 +106,58 @@
...
+                environment.systemPackages = [ pkgs.jq ];
+                # environment.variables =
+                #   builtins.readDir ./test-keys
+                #   |> builtins.attrNames
+                #   |> builtins.filter (x: x == lib.toUpper x)
+                #   |> builtins.map (x: {
+                #     name = x;
+                #     value = "x";
+                #   })
+                #   |> builtins.listToAttrs;
```
reo101, line 155: This looks good, why is it commented out?

nix/test-environments/default.nix
```diff
@@ -102,5 +106,58 @@
...
+                  };
+                };
+
+                security.pam.services.sshd.allowNullPassword = true;
+                networking.nameservers = [ "8.8.8.8" ];
+                environment.systemPackages = [ pkgs.jq ];
+                # environment.variables =
+                #   builtins.readDir ./test-keys
+                #   |> builtins.attrNames
+                #   |> builtins.filter (x: x == lib.toUpper x)
```
reo101, line 150: Can probably match `x` against an all-uppercase regex for a small perf boost

nix/test-environments/test-setup.nix
```diff
@@ -0,0 +1,35 @@
...
+      a = {
+        api-keys = lib.mkForce {
+          ALPHAVANTAGE_API_KEY = "${testKeysDir}/ALPHAVANTAGE_API_KEY";
+          APCA_API_KEY_ID = "${testKeysDir}/APCA_API_KEY_ID";
+          APCA_API_SECRET_KEY = "${testKeysDir}/APCA_API_SECRET_KEY";
+          YAHOO_FINANCE_API_KEY = "${testKeysDir}/YAHOO_FINANCE_API_KEY";
+          TWELVEDATA_API_KEY = "${testKeysDir}/TWELVEDATA_API_KEY";
+          FMP_API_KEY = "${testKeysDir}/FMP_API_KEY";
+          SPOUT_RWA_API_KEY = "${testKeysDir}/SPOUT_RWA_API_KEY";
+        };
```
reo101, line 30: This looks trivially automatable

.envrc
```diff
@@ -11,6 +11,10 @@ dotenv_if_exists
 mkdir -p "$PWD/.devenv"
 DEVENV_ROOT_FILE="$PWD/.devenv/root"
 printf %s "$PWD" > "$DEVENV_ROOT_FILE"
+
+git update-index --assume-unchanged nix/test-environments/test-keys/[A-Z]*
+git update-index --skip-worktree nix/test-environments/test-keys/[A-Z]*
```
reo101, line 16: Would benefit from a comment explaining the need and purpose for those two commands, given their peculiarity

Justfile
```diff
@@ -138,6 +138,37 @@ start-oracle oracle-name trigger-oracle-build-type="--use-local-cargo-artifacts"
...
+
+[group('General')]
+[doc('Run NixOS integration tests')]
+run-nixos-tests:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    tests=$(just list-nixos-tests)
+    for test in $tests; do
+      echo "Running NixOS test: $test"
+      nix build --option sandbox false -L --json --accept-flake-config .#legacyPackages.x86_64-linux.nixosTests."$test"
```
reo101, line 169: Why the disabled sandbox?

nix/test-environments/test-script.py
```diff
@@ -0,0 +1,300 @@
...
+import json
+import time
+from dataclasses import dataclass
+from typing import List
+from test_driver.errors import RequestedAssertionFailed # type: ignore
+
+# --- Configuration ---
+@dataclass
+class TestConfig:
+    """Centralized configuration for the integration test."""
```
reo101, line 12: Quite a lot of constants here, would it not be better to provide them from the `nix` side? Not sure what's the preferred way when writing `nixos` integration tests, might be perfectly fine this way

nix/test-environments/test-script.py
```diff
@@ -0,0 +1,300 @@
...
+    UPGRADEABLE_PROXY_CONTRACT: str = "0xee5a4826068c5326a7f06fd6c7cbf816f096846c"
+    CHAINLINK_PROXY_CONTRACT: str = "0x9fAb38E38d526c6ba82879c6bDe1c4Fd73378f17"
+    UPGRADEABLE_PROXY_ADFS_CONTRACT: str = "0xADF5aad6faA8f2bFD388D38434Fc45625Ebd9d3b"
+    CL_AGGREGATOR_ADAPTER_CONTRACT: str = "0xcBD6FC059bEDc859d43994F5C221d96F9eD5340f"
+
+    # Service Names
+    REPORTER_SERVICE: str = "blocksense-reporter-a.service"
+    ANVIL_SEPOLIA_SERVICE: str = "blocksense-anvil-ethereum-sepolia.service"
+    ANVIL_INK_SERVICE: str = "blocksense-anvil-ink-sepolia.service"
+    SEQUENCER_SERVICE: str = "blocksense-sequencer.service"
```
reo101, line 28: Might want to make this more dynamic (at least the `-a` part), currently hard-dependening on the nix config outside:

https://github.com/blocksense-network/blocksense/pull/1279/files#diff-f195c6435e680fa841aab93a92956dc6cbc0dce5f9ff63ece2c68e31979ffbc0R21

nix/test-environments/test-script.py
```diff
@@ -0,0 +1,300 @@
...
+        "get_history",
+        "get_oracle_scripts"
+    ]
+
+    for endpoint in json_endpoints:
+        check_sequencer_endpoint_json(endpoint)
+        print(f"✓ {endpoint} endpoint is responding with valid JSON")
+
+    # Non-JSON endpoints
+    # check_metrics_endpoint() # metrics aren't wokring
```
reo101, line 145: Should we not check this out (i.e. what's wrong with it)?

nix/test-environments/example-setup-03.nix
```diff
@@ -91,15 +91,14 @@ in
...
-      cex-price-feeds.exec-interval = lib.mkForce 10;
-      exsat-holdings.exec-interval = lib.mkForce 10;
-      gecko-terminal.exec-interval = lib.mkForce 10;
-      eth-rpc.exec-interval = lib.mkForce 10;
-      stock-price-feeds.exec-interval = lib.mkForce 10;
+      cex-price-feeds.exec-interval = lib.mkForce 60;
+      exsat-holdings.exec-interval = lib.mkForce 60;
+      gecko-terminal.exec-interval = lib.mkForce 60;
+      eth-rpc.exec-interval = lib.mkForce 60;
+      stock-price-feeds.exec-interval = lib.mkForce 60;
```
reo101, line 101: Why the change to `60` for all of those? Might be residue from rebasing?

nix/modules/blocksense/backends/process-compose.nix
```diff
@@ -189,13 +189,14 @@ in
...
+    services.kafka.enable = cfg.kafka.enable;
     processes = lib.mkMerge [
       anvilImpersonateAndFundInstances
       reporterInstances
       sequencerInstance
       anvilInstances
-      (lib.mkIf config.services.kafka.enable blockchainReader)
-      (lib.mkIf config.services.kafka.enable aggregateConsensusReader)
+      (lib.mkIf cfg.kafka.enable blockchainReader)
+      (lib.mkIf cfg.kafka.enable aggregateConsensusReader)
```
reo101, line 199: Good :D

.envrc
reo101: Might also want to define a more generic way to have some files *ignored* by `git`, at least in most of the meaningful ways

nix/test-environments/test-setup.nix
```diff
@@ -0,0 +1,35 @@
+{
+  lib,
+  ...
+}:
+let
+
+  root = ../..;
+
+  testKeysDir = lib.path.append root "./nix/test-environments/test-keys";
+
```
reo101, line 10: nit:

```suggestion
  root = ../..;

  testKeysDir = lib.path.append root "./nix/test-environments/test-keys";
```

nix/test-environments/test-script.py
reo101, line 1: All those `# type: ignore`s are a bit icky, is there something better we can do about them?

nix/modules/blocksense/module-opts/default.nix
```diff
@@ -19,6 +19,10 @@ in
     Enable the Blocksense sequencer and reporter node services.
   '');
 
+  kafka.enable = mkEnableOption (mdDoc ''
+    Enable Kafka.
+  '');
```
reo101, line 24: Why is this inside the `blocksense` *module* attribute now? Not saying that it shouldn't, just asking why.

On a similar note, why is `blama` inside too? Since it (or they?) can be useful outside the whole of `blocksense`, we might leave it out (leaving the possibility of writing a test someday just for `blama` alone)

nix/machines/default.nix
```diff
@@ -0,0 +1,68 @@
+{
+  self,
+  lib,
+  ...
+}:
+let
+  allEnvironmentPaths = lib.filter (path: lib.hasPrefix "example-setup-" (builtins.baseNameOf path)) (
+    lib.attrValues (builtins.readDir ../test-environments)
+  );
```
reo101, line 9: That's a mouthful, might benefit from a `lib.pipe` or a `|>`-chain

nix/test-environments/test-script.py
```diff
@@ -0,0 +1,300 @@
...
+    safe_rpc_url = shlex.quote(rpc_url)
+    network_name = _get_network_name_from_rpc_url(rpc_url)
+
+    if re.match(r"^0x[0-9a-fA-F]+$", call_data):
+        call_data_arg = f"--data {shlex.quote(call_data)}"
+    else:
+        call_data_arg = shlex.quote(call_data)
+
+    # Execute just the cast call without post-processing pipeline
+    command = f"cast call {safe_contract_address} {call_data_arg} --rpc-url {safe_rpc_url}"
```
reo101, line 174: Aha, this raw `cast` call is maybe why we need `foundry` in the `systemPackages` for the test machine, fair. Maybe comment that better here and/or there (here - the assumption that those binaries are available because of prior setup, there - the reason for the addition of the package)

nix/test-environments/test-script.py
```diff
@@ -0,0 +1,300 @@
...
+    # Using CLAggregatorAdapter contract - calls latestAnswer()
+    make_cast_call_ink(test_config.CL_AGGREGATOR_ADAPTER_CONTRACT)
+
+    print("✓ All contracts responded successfully")
+
+
+def wait_for_sequencer_history():
+    """Waits until the sequencer history endpoint reports non-empty data."""
+    machine.wait_until_succeeds(  # type: ignore
+        f"curl -s {test_config.SEQUENCER_URL}/get_history | jq -e '.aggregate_history | all(. != [])'",
```
reo101, line 245: Hmmm, `curl` and `jq` calls too, did I miss where we add those to the env?

nix/test-environments/test-script.py
```diff
@@ -0,0 +1,300 @@
...
+    def ports_to_wait_for(self) -> List[int]:
+        return [self.SEPOLIA_PORT, self.INK_PORT, self.SEQUENCER_PORT]
+
+# Create a single instance of the config to be used throughout the script
+test_config = TestConfig()
+
+
+def wait_for_services(services):
+    for service in services:
+        machine.wait_for_unit(service)  # type: ignore
```
reo101, line 68: Global usage of `machine` seems icky, could we maybe substitute it for a more descriptive name? Maybe also somehow, somewhere declare it's actual type, so that the automatic type-checker does not yell at us?

nix/test-environments/test-script.py
```diff
@@ -0,0 +1,300 @@
...
+def wait_for_ports(ports):
+    for port in ports:
+        machine.wait_for_open_port(port)  # type: ignore
+
+
+def check_log_for_pattern(service, pattern):
+    """Asserts that a pattern is NOT found in a service's journald log."""
+    log_output = machine.succeed(f"journalctl -u {service}")  # type: ignore
+
+    # Extract logic into Python with better error handling
```
reo101, line 80: Could maybe clean up some of the leftover LLM comments (or make them *official* and spread them across the whole script, adding context to more things)

nix/test-environments/test-script.py
reo101, line 1: We should maybe look into `pre-commit` hooks for formatting and linting Python code

nix/machines/default.nix
PetarKirov: Agreed, though if we test solely `x86_64-linux`, then perhaps the extra complexity is not worth it.

nix/machines/default.nix
```diff
@@ -0,0 +1,68 @@
...
+      fsType = "ext4";
+    };
+    nix.settings.sandbox = false;
+    networking.useNetworkd = true;
+    systemd = {
+      network.enable = true;
+      network.wait-online.enable = false;
+      services.network-online.enable = true;
+    };
+    environment.defaultPackages = [ self.legacyPackages.x86_64-linux.foundry ];
```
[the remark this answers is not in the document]
PetarKirov: It's used by `nix/test-environments/test-script.py`, e.g. the `cast` command.

nix/test-environments/default.nix
```diff
@@ -102,5 +106,58 @@
           ))
           lib.listToAttrs
         ];
+      packages = {
+        inherit (self'.legacyPackages.process-compose-environments.with-local-cargo-artifacts) all;
+      };
+
+      legacyPackages.nixosTests = {
+        # example-setup-03 is faster than example-setup-01, so we use it as the default test environment.
```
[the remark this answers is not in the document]
PetarKirov: Yes, though this is off-topic for this PR.

nix/test-environments/default.nix
PetarKirov: We support older version of Nix (say up to 2.18.x) where the pipe operators are either not enabled by default, or not available in the first place.

Justfile
PetarKirov: Most likely to enable internet access, needed for both `anvil --fork-url` and the oracle scripts.

nix/test-environments/test-script.py
PetarKirov: I wonder where we can take them from the generated by nix config files (sequencer and reporter config json)

nix/machines/default.nix
PetarKirov: we're not leaving in the nix future yet

PR #1281, opened 2025-05-02
feat(oracles): Add RPC Etherium oracle YToken echangeRate

yordanmadzhunkov:
Added yield nest price feeds:

   1. report YieldFiyUSD -> exchangeRate from smart contract address https://etherscan.io/token/0x2799dE2E1e769fA58dd3787F70BcD839AF3a1F39#readProxyContract

   2. Report YnBNBx from BNB chain. This requires different RPCs

   3. Added a fallback mechanism in case one RPC does not return useful data

```bash
+---------+---------------------------+----------------------------------+---------------------------------------------+
| Feed ID |          Method           |              Value               |                   RPC Url                   |
+---------+---------------------------+----------------------------------+---------------------------------------------+
|  100001 | YnETHx -> convertToAssets | Numerical(1.0451593123791621e18) | Some("https://ethereum-rpc.publicnode.com") |
|  100002 | YToken -> exchangeRate    |             Numerical(1083770.0) |          Some("https://rpc.eth.gateway.fm") |
|  100003 | YnBNBx -> convertToAssets | Numerical(1.0246939620277382e18) |        Some("https://binance.llamarpc.com") |
+---------+---------------------------+----------------------------------+---------------------------------------------+
```

### DIFF
--- a/apps/oracles/eth-rpc/src/lib.rs
+++ b/apps/oracles/eth-rpc/src/lib.rs
@@ -12,6 +12,7 @@ use blocksense_sdk::{
 };
 use prettytable::{format, Cell, Row, Table};
 use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
 use std::str::FromStr;
 use url::Url;
 
@@ -22,6 +23,14 @@ sol!(
     #[sol(rpc)]
     YnETHx,
     "src/abi/YnETHx.json"
+);
+
+sol!(
+    #[allow(missing_docs)]
+    #[allow(clippy::too_many_arguments)]
+    #[sol(rpc)]
+    YieldFiyUSD,
+    "src/abi/YieldFiyUSD.json"
 );
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -51,6 +60,14 @@ pub struct ResponseEthCall {
     pub id: Option<u64>,
     pub error: Option<ResponseEthCallError>,
     pub result: Option<String>,
+    #[serde(default)]
+    pub rpc_url: Option<String>,
+}
+
+#[derive(Deserialize, Debug, Clone)]
+pub struct ResponsePerFeed {
+    pub feed_id: u128,
+    pub res: ResponseEthCall,
 }
 
 fn convert(x: [u8; 32]) -> ([u8; 16], [u8; 16]) {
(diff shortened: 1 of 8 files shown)


apps/oracles/eth-rpc/src/lib.rs
```diff
@@ -109,54 +126,145 @@ impl RequestEthCall {
...
-    let rpc_url: Url = "https://eth.llamarpc.com".parse()?;
-    let yn_eth_x_contract_address = address!("0x657d9ABA1DBb59e53f9F3eCAA878447dCfC96dCb");
+type FetchedDataForFeed = HashMap<u64, ResponseEthCall>;
 
-    // Create a provider, in order to create instance of the contract
-    let provider = ProviderBuilder::new().on_http(rpc_url.clone());
-    let contact = YnETHx::new(yn_eth_x_contract_address, provider);
-    let shares = U256::from(1_000_000_000_000_000_000_u64);
+async fn fetch_all(resourses: &[FeedConfig]) -> Result<FetchedDataForFeed> {
+    //let rpc_url = "https://rpc.eth.gateway.fm".parse()?;
```
melatron, line 132: nit: delete

apps/oracles/eth-rpc/src/lib.rs
```diff
@@ -51,6 +60,14 @@ pub struct ResponseEthCall {
     pub id: Option<u64>,
     pub error: Option<ResponseEthCallError>,
     pub result: Option<String>,
+    #[serde(default)]
+    pub rpc_url: Option<String>,
+}
+
+#[derive(Deserialize, Debug, Clone)]
+pub struct ResponsePerFeed {
+    pub feed_id: u64,
```
melatron, line 69: nit: should be u128 as per - https://github.com/blocksense-network/blocksense/pull/1358

melatron (dismissed):

apps/oracles/eth-rpc/src/lib.rs
```diff
@@ -109,54 +126,145 @@ impl RequestEthCall {
...
+                }
+                if final_value.is_none() {
+                    final_value = Some(value);
+                }
+            }
+            if let Some(value) = final_value {
+                res.insert(feed_config.feed_id, value);
+            }
+        }
+    }
```
EmilIvanichkovv, line 216: I believe we've already discussed and agreed that we don’t want to introduce any conditional logic in the oracle based on specific feed IDs. The correct approach is to keep everything feed-specific within the feed’s `arguments` field. The oracle should execute the appropriate workflow based on the contents of that field — not based on hardcoded ID checks.

What’s more, there’s a lot of duplicated code in this part, which we could avoid by introducing a dedicated function to handle this logic — extracting the necessary data from the `arguments` field and running the appropriate workflow accordingly.

config/feeds_config_v2.json
```diff
@@ -52880,7 +52880,68 @@
...
+      "additional_feed_info": {
+        "pair": {
+          "base": "",
+          "quote": ""
+        },
+        "decimals": 0,
+        "category": "",
+        "market_hours": "Crypto",
+        "arguments": [],
+        "rpc_urls": ["https://eth.llamarpc.com", "https://ethereum-rpc.publicnode.com", "https://rpc.eth.gateway.fm"]
```
EmilIvanichkovv, line 52914: If you still decide to go with this approach, make sure to update the feed config schema in `libs/ts/config-types/src/data-feeds-config`. Otherwise, we risk losing the `rpc_urls` field the next time we auto-generate `config/feeds_config_v2.json`.

EmilIvanichkovv (commented):
I have some concerns on merging this.

apps/oracles/eth-rpc/src/lib.rs
yordanmadzhunkov: I agree, but note that different smart contracts are called with different number of arguments. I will reduce the code duplication, but the approach based on feed id stays for now.

config/feeds_config_v2.json
```diff
@@ -52880,7 +52880,68 @@
         "decimals": 0,
         "category": "",
         "market_hours": "Crypto",
-        "arguments": []
+        "arguments": [],
+        "rpc_urls": ["https://eth.llamarpc.com", "https://ethereum-rpc.publicnode.com", "https://rpc.eth.gateway.fm"]
```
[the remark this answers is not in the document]
yordanmadzhunkov: It's better to use `arguments` moving data there!

apps/oracles/eth-rpc/src/lib.rs
```diff
@@ -109,54 +126,129 @@ impl RequestEthCall {
...
+    let mut res = FetchedDataForFeed::new();
+    for feed_config in resourses {
+        let mut final_value = None;
+        for rpc_url_candidate in &feed_config.arguments {
+            let rpc_url: Url = rpc_url_candidate.as_str().parse()?;
+            let provider = ProviderBuilder::new().on_http(rpc_url.clone());
+            let eth_call = match feed_config.feed_id {
+                100001 => yn_ethx_call(provider.clone(), 12345 + 998),
+                100002 => yield_fi_yusd_call(provider.clone(), 12345 + 128),
+                100003 => yn_bnbx_call(provider.clone(), 12345 + 129),
```
EmilIvanichkovv, line 183: I like this more.

apps/oracles/eth-rpc/src/lib.rs
EmilIvanichkovv: I see. I am ok with your change

config/feeds_config_v2.json
EmilIvanichkovv: I understand, but this could lead to bigger issues down the line. Whenever we make changes to the oracles (and, by extension, the feeds config), we need to ensure the schemas are kept in sync. Otherwise, we risk inconsistencies that are hard to track and debug later.

I added a commit to your branch that defines the `arguments` for the `eth-rps` oracle. The build should now pass.

EmilIvanichkovv (approved):

EmilIvanichkovv (approved):

PR #1289, opened 2025-05-07
Use cargo in devenv process compose

Xearty:


HristoStaykov (dismissed):

Justfile
```diff
@@ -33,6 +33,16 @@ start-oracle oracle-name:
   cd "{{root-dir}}/apps/oracles/{{oracle-name}}"
   RUST_LOG=trigger=info "${SPIN:-spin}" build --up
 
+build-blocksense:
+  @{{root-dir}}/scripts/build-blocksense.sh
+
+start-blocksense:
+  #!/usr/bin/env bash
```
reo101, line 40: Might prefer `/bin/sh` here if `set -euo pipefail` is POSIX

reo101 (approved):

nix/modules/blocksense/backends/process-compose.nix
```diff
@@ -1,17 +1,21 @@
...
 }:
 let
   inherit (self'.apps)
-    sequencer
-    blocksense
     blockchain_reader
     aggregate_consensus_reader
     ;
 
+  mkCargoTargetExePath = executable-name: "${config.devenv.root}/target/release/${executable-name}";
```
EmilIvanichkovv, line 17: How do I produce the release build?
cargo build --release ?

For example, I have run just clean and lost my target folder. And I expect direnv reload & process-compose to do the work. But I see this log from the sequencer

```
bash: line 1: /home/emilivanichkov/code/repos/metacraft-labs/blocksense/monorepo/target/release/sequencer: No such file or directory
```

In the PR description, there are no instruction for the new workflow, nor there are such in our discord. Please provide

PR #1327, opened 2025-05-13
[nft/social-verification]: Simplify form and flow

RadoslavDimchev:
Removed the third step to retweet (moved to success form) and removed check for verifying retweet with unique code (replaced code from signing with keccak256)


RadoslavDimchev:
#1328

PR #1348, opened 2025-05-17
refactor(social-verification): Use `participants_unique` DB

EmilIvanichkovv:
Also add the db schema


PetarKirov (approved):

apps/social-verification/src/db-schema.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE
```
DannyStoyanov, line 1: nit: I am not completely sure how Cloudflare D1 works but in general whenever you have to create a new table a better practice is to use `CREATE TABLE IF NOT EXISTS` to prevent errors like "table already exists" unless you have strong reason to skip the `...IF NOT EXISTS` part.

PR #1355, opened 2025-05-19
fix(spin-plugin): Dynamically discover the host os and arch when installing plugin

Xearty:


scripts/get-host-arch-and-os.sh
EmilIvanichkovv, line 1: This looks fine to me
image

EmilIvanichkovv (commented):
I still can build.
image


Lets see results on mac

EmilIvanichkovv (commented):
On a clean clone I see this error when I run `just start-oracle <oracle-name>`

```
Error: metadata error: invalid metadata value for "triggers": Error("unknown field `kafka_endpoint`, expected one of `interval_time_in_seconds`, `sequencer`, `prometheus_url`, `secret_key`, `reporter_id`", line: 0, column: 0)
error: Recipe `start-oracle` failed with exit code 1
```

In my every-day-work repo I do not have a problem with the command.. 
Noticed that I have different versions of spin..
- every-day-work repo
image
- new clone 
image


Do you think this might cause the problem?

PR #1357, opened 2025-05-19
fix(just): Fix `start-oracle` command

Xearty:
-- export `SPIN_DATA_DIR`


EmilIvanichkovv (approved):
Ahh nice one.  Tested, it is working!
Thanks

DannyStoyanov:
Tested on MacOS too (because I had different error) - works as expected!

PR #1358, opened 2025-05-19
Hstaykov/feed id to 128b

HristoStaykov:
Refactor Feed ID type to be u128 and use type alias.


apps/sequencer/src/providers/provider.rs
```diff
@@ -369,19 +370,20 @@ impl RpcProvider {
...
-        feed_ids: &[u32],
+        feed_ids: &[FeedId],
     ) -> Result<Vec<Result<PublishedFeedUpdate, PublishedFeedUpdateError>>, eyre::Error> {
         let multicall = self.get_contract_address(MULTICALL_CONTRACT_NAME)?;
         let data_feed = self.get_contract_address(PRICE_FEED_CONTRACT_NAME)?;
         let contract = MulticallInstance::new(multicall, self.provider.clone());
         let calldata: Vec<Multicall::Call> = feed_ids
             .iter()
             .map(|feed_id| {
+                let feed_id = *feed_id as u32;
```
melatron, line 381: What if it is bigger then u32? Isn't this the reason we are moving from u32?

melatron (changes requested):
I don't like that we are parsing FeedId which is u128 to u32. Can you please write a test witch feedId = `u128::MAX`.

apps/sequencer/src/providers/provider.rs
HristoStaykov: I think this functionality for reading from the contract is only relevant to the old version of the contracts. In the old version of the contracts feed id-s are 4 bytes. This is also why when we call legacy serialize we cast to u32 as well.

melatron: @HristoStaykov said he will refactor the Sequencer test to work with ADFS and will test for u128::MAX in another PR.

melatron (approved):

HristoStaykov:
> u128::MAX

We need to refactor the integration tests to use ADFS version of the contract to be able to implement a thorough test with feed id = u128::MAX

melatron (dismissed):

melatron (approved):

PR #1364, opened 2025-05-21
fix(sequencer/eth_send_utils): Log block height on generation of call data

HristoStaykov:


yordanmadzhunkov (approved):

apps/sequencer/src/providers/eth_send_utils.rs
```diff
@@ -618,11 +619,12 @@ pub async fn eth_batch_send_to_all_contracts(
 
     let result = futures::future::join_all(collected_futures).await;
     let mut all_results = String::new();
+    let block_height = updates.block_height;
     for v in result {
         match v {
             Ok((result, net, provider)) => match result.await {
                 Ok((status, updated_feeds)) => {
-                    all_results += &format!("result from network {net}: Ok -> status: {status}");
+                    all_results += &format!("result from network {net} and block_height {block_height}: Ok -> status: {status}");
```
Owliie, line 627: Either log `block_height` or `block height` but not both

Owliie (changes requested):

apps/sequencer/src/providers/eth_send_utils.rs
Owliie: This here is still not resolved

PR #1365, opened 2025-05-22
feat(sequencer/metrics): Add heartbeat to reports metrics dimensions

HristoStaykov:


apps/sequencer/src/providers/eth_send_utils.rs
```diff
@@ -748,7 +748,7 @@ async fn increment_feeds_round_metrics(
     if let Some(ref fm) = feeds_metrics {
         for feed in updated_feeds {
             // update the round counters' metrics accordingly
-            inc_vec_metric!(fm, feed, updates_to_networks, net);
```
yordanmadzhunkov, line 751: interesting, it this a bug fix ?

yordanmadzhunkov: I saw the change in the macro, now it makes sense to me. Never mind.

yordanmadzhunkov (approved):
Cool

PR #1367, opened 2025-05-23
fix(sequencer/metrics): Put feed name as additional info into metrics

HristoStaykov:


apps/sequencer/src/feeds/feed_slots_processor.rs
```diff
@@ -192,8 +196,7 @@ impl FeedSlotsProcessor {
 
         if skip_decision.should_skip() {
             info!(
-                "Skipping publishing for feed_id = {} change is lower then threshold of {} %",
-                feed_id, skip_publish_if_less_then_percentage
+                "Skipping publishing for [feed `{feed_name}` feed_id = {feed_id}] change is lower then threshold of {skip_publish_if_less_then_percentage} %",
```
melatron, line 199: is this comma at the end needed?

melatron (approved):
LGTM with small nit

PR #1386, opened 2025-06-02
refactor(sequencer/providers): Check in parallel for all contracts existence on all networks during startup

HristoStaykov:


apps/sequencer/src/providers/provider.rs
```diff
@@ -169,6 +169,16 @@ async fn get_rpc_providers(
...
 }
 
+async fn log_if_contracts_exist(provider_mutex: Arc<Mutex<RpcProvider>>) {
+    let rpc_provider = provider_mutex.lock().await;
+    rpc_provider
+        .log_if_contract_exists(PRICE_FEED_CONTRACT_NAME)
+        .await;
+    rpc_provider
+        .log_if_contract_exists(EVENT_FEED_CONTRACT_NAME)
+        .await;
```
reo101, line 179: Have you considered doing these two in parallel too?

PR #1395, opened 2025-06-05
docs(ui): document the Blocksense UI library and its publishing process

zah:


Justfile
```diff
@@ -56,3 +56,27 @@ clean:
...
+
+  # Publish
+  npm publish
+
+  # Create git tag
+  git add package.json
+  git commit -m "chore: release @blocksense/ui@$(node -p "require('./package.json').version")"
+  git tag "ui-v$(node -p "require('./package.json').version")"
+
+  echo "✅ Published @blocksense/ui@$(node -p "require('./package.json').version")"
```
EmilIvanichkovv, line 82: But we do not publish..

EmilIvanichkovv (commented):
We are not there yet.. I prefer to close this PR as most of the information in it is inaccurate.

It would be a lot time-consuming to make a proper review and only leave the correct info

libs/ts/ui/README.md
EmilIvanichkovv, line 1: I see valuable info here, tho not everything is correct.

PR #1396, opened 2025-06-05
fix(sequencer/eth_send_utils): Name all the spawned sender tasks and await them correctly

HristoStaykov:


apps/sequencer/src/feeds/votes_result_sender.rs
```diff
@@ -60,15 +60,18 @@ pub async fn votes_result_sender_loop(
...
 ) {
+    let blocksense_block_height = updates.block_height;
     let sender = tokio::task::Builder::new()
-        .name(format!("batch_sender_{batch_count}").as_str())
+        .name(
+            format!("async_send_to_all_networks_{blocksense_block_height}_{batch_count}").as_str(),
+        )
         .spawn_local(async move {
-            debug!("Spawned batch_sender_{batch_count}");
+            debug!("Spawned async_send_to_all_networks_{batch_count}");
```
PetarKirov, line 74: Add block height here as well

PetarKirov (approved):
Nice work!

PR #1399, opened 2025-06-05
feat(devshell): print out the most useful just commands when you enter the devshell

zah:


nix/shells/pkg-sets/dev-shell.nix
```diff
@@ -44,5 +44,24 @@
...
+      echo ""
+      echo "🛠️  Development commands:"
+      echo "  🎨 just format               - Format all code (JS/Rust/Nix)"
+      echo "  🔍 just lint                 - Run all linters"
+      echo "  🔧 just fix-all              - Format and fix all linting issues"
+      echo "  ✅ just check-all            - Run all checks"
+      echo "  🧹 just clean                - Clean build artifacts"
+      echo ""
+      echo "💡 Run 'just' to see all available targets"
+      echo ""
```
EmilIvanichkovv, line 65: It would be nice to be auto-generated, not hard-coded, as we will extend the just commands.

btw `just` have it
image


@PetarKirov showed me that we can also group the common recepies with [this feature](https://just.systems/man/en/groups.html#groups): e.g:
```
[group('lint')]
js-lint:
    echo 'Running JS linter…'

[group('rust recipes')]
[group('lint')]
rust-lint:
    echo 'Running Rust linter…'

[group('lint')]
cpp-lint:
  echo 'Running C++ linter…'

# not in any group
email-everyone:
    echo 'Sending mass email…'
```

PR #1405, opened 2025-06-09
Fix/erroneous logging fail tx

HristoStaykov:
Refactor to fix frequently timing out get_receipt from RPC provider.


apps/sequencer/src/providers/eth_send_utils.rs
```diff
@@ -272,32 +275,78 @@ pub async fn eth_batch_send_to_contract(
...
+            &sender_address,
+            block_height,
+            transaction_retry_timeout_secs,
+            false,
+        )
+        .await
+        {
+            Ok(latest_nonce) => {
+                if latest_nonce > nonce {
+                    //TODO: Check the tx hashes of all posted/retried txs for block inclusion
```
Owliie, line 350: Is this left for another PR?

PR #1417, opened 2025-06-12
fix(direnv): Fix direnv on macOS systems with old version of `find`

PetarKirov:
The default version of find (`/usr/bin/find`) doesn't support `-p`.
Additionally the `watch_file` function sometimes is not able to accept long list of files.


EmilIvanichkovv (commented):
lame

PR #1424, opened 2025-06-17
Hstaykov/process txs parallel over networks through channels

HristoStaykov:
In this PR we create a separate pipeline per network. Transactions are processed in parallel between pipelines, but sequential in each pipeline + retry/drop mechanisms are preserved. This change in combination with configuration specified per network feeds allows high frequency updates for those feeds to networks with such requirements.

### DIFF
--- a/apps/sequencer/src/providers/eth_send_utils.rs
+++ b/apps/sequencer/src/providers/eth_send_utils.rs
@@ -261,9 +261,10 @@ pub async fn loop_processing_batch_of_updates(
 ) {
     tracing::info!("Starting {relayer_name} loop...");
 
+    //TODO: Create a termination reason pattern in the future. At this point networks are not added/removed dynamically in the sequencer,
+    // therefore the loop in iterating over the lifetime of the sequencer.
     loop {
         let cmd_opt = chan.recv().await;
-        // tracing::info!("{relayer_name} received {cmd:?}");
         match cmd_opt {
             Some(cmd) => {
                 let block_height = cmd.updates.block_height;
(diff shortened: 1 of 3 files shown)


apps/sequencer/src/providers/eth_send_utils.rs
```diff
@@ -205,6 +211,133 @@ pub async fn get_serialized_updates_for_network(
...
+    relayer_name: String,
+    feeds_metrics: Arc<RwLock<FeedsMetrics>>,
+    provider_status: Arc<RwLock<HashMap<String, ProviderStatus>>>,
+    mut chan: UnboundedReceiver<BatchOfUpdatesToProcess>,
+) {
+    tracing::info!("Starting {relayer_name} loop...");
+
+    loop {
+        let cmd_opt = chan.recv().await;
+        // tracing::info!("{relayer_name} received {cmd:?}");
```
melatron, line 266: Remove

apps/sequencer/src/providers/eth_send_utils.rs
```diff
@@ -205,6 +211,133 @@ pub async fn get_serialized_updates_for_network(
...
+pub async fn loop_processing_batch_of_updates(
+    net: String,
+    relayer_name: String,
+    feeds_metrics: Arc<RwLock<FeedsMetrics>>,
+    provider_status: Arc<RwLock<HashMap<String, ProviderStatus>>>,
+    mut chan: UnboundedReceiver<BatchOfUpdatesToProcess>,
+) {
+    tracing::info!("Starting {relayer_name} loop...");
+
+    loop {
```
melatron, line 264: This loop is never ending. You can add a TODO to create a `termination reason` pattern in the future. This way we will have more control for the workers.

melatron (dismissed):
Consider using broadcast instead of 1:1 channels. Maybe in the future we will need 1:1 channels for per network messages, but currently we are kind of broadcasting.

PR #1429, opened 2025-06-19
[changelog-generator]: Add `list feeds` command

DannyStoyanov:
### Description
This is a preview how to add simple command within `Effect CLI app` and how to run the command.

### Usage
Run the following command in the `/apps/changelog-generator/`:

```
yarn start list-feeds --network <network>
```
where `<network>` is for example: `ink-sepolia`, `citrea-testnet`, etc.

As a result you should see a table containing `feed name`, `address`, `decimals`, `base` and `quote` for each available feed. 

The output of running the above mentioned command is:
Screenshot 2025-06-20 at 13 16 39

### Wizard mode 🧙🏻‍♂️ 
Additionally you can run
```
yarn start list-feeds --wizard
```
where you enter the interactive mode of the CLI app. The result is shown below:

[asciicast](https://asciinema.org/a/0eDUR1Q9ci77CagSUPpOFClkl)

### Additional info
Apart from that, in the `/apps/changelog-generator/` you can run the following command for more info:
```
yarn node ./dist/bin.cjs
```

Screenshot 2025-06-19 at 16 44 11

### DIFF
--- a/apps/changelog-generator/src/Cli.ts
+++ b/apps/changelog-generator/src/Cli.ts
@@ -1,5 +1,7 @@
 import { Command } from '@effect/cli';
 
+import packageJson from '../package.json';
+
 import { listFeeds } from './commands/list-feeds';
 
 const command = Command.make('changelog-generator').pipe(
@@ -7,6 +9,6 @@ const command = Command.make('changelog-generator').pipe(
 );
 
 export const run = Command.run(command, {
-  name: 'Changelog Generator',
-  version: '0.1.0',
+  name: packageJson.description,
+  version: packageJson.version,
 });
--- a/apps/changelog-generator/src/Cli.ts
+++ b/apps/changelog-generator/src/Cli.ts
@@ -1,8 +1,12 @@
-import * as Command from '@effect/cli/Command';
+import { Command } from '@effect/cli';
 
-const command = Command.make('hello');
+import { listFeeds } from './commands/list-feeds';
+
+const command = Command.make('changelog-generator').pipe(
+  Command.withSubcommands([listFeeds]),
+);
 
 export const run = Command.run(command, {
-  name: 'Hello World',
-  version: '0.0.0',
+  name: 'Changelog Generator',
+  version: '0.1.0',
 });
--- a/Justfile
+++ b/Justfile
@@ -63,6 +63,7 @@ build-ts package="all":
     yarn build-single @blocksense/sol-reflector
     yarn build-single @blocksense/contracts
     yarn build-single @blocksense/data-feeds-config-generator
+    yarn build-single @blocksense/changelog-generator
   else
     yarn build:recursive {{package}}
   fi
--- a/apps/changelog-generator/package.json
+++ b/apps/changelog-generator/package.json
@@ -18,6 +18,8 @@
     "changeset-publish": "yarn build && TEST_DIST= yarn vitest && changeset publish"
   },
   "dependencies": {
+    "@blocksense/base-utils": "workspace:^",
+    "@blocksense/config-types": "workspace:^",
     "@effect/cli": "0.63.9",
     "@effect/cluster": "^0.38.14",
     "@effect/experimental": "^0.48.10",
--- a/apps/changelog-generator/package.json
+++ b/apps/changelog-generator/package.json
@@ -7,7 +7,7 @@
   "bin": "./dist/bin.cjs",
   "scripts": {
     "build": "tsup --dts && yarn copy-package-json",
-    "ts": "yarn node --import tsx",
+    "start": "yarn node --import tsx ./src/bin.ts",
     "build:ts": "tsup",
     "lint": "yarn run -T eslint \"**/{src,test,examples,scripts,dtslint}/**/*.{ts,mjs}\"",
     "lint-fix": "yarn lint --fix",
--- a/apps/changelog-generator/package.json
+++ b/apps/changelog-generator/package.json
@@ -1,9 +1,9 @@
 {
   "name": "@blocksense/changelog-generator",
-  "version": "0.0.0",
+  "version": "0.1.0",
   "type": "module",
   "license": "MIT",
-  "description": "An Effect app related with Release notes functionality",
+  "description": "Blocksense Changelog Generator",
   "bin": "./dist/bin.cjs",
   "scripts": {
     "build": "tsup --dts && yarn copy-package-json",
--- a/apps/changelog-generator/src/bin.ts
+++ b/apps/changelog-generator/src/bin.ts
@@ -7,5 +7,5 @@ import { run } from './Cli.js';
 
 run(process.argv).pipe(
   Effect.provide(NodeContext.layer),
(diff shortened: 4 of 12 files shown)


apps/changelog-generator/src/Cli.ts
```diff
@@ -1,8 +1,108 @@
...
+        contract?.quote ?? null,
+      ]);
+    });
+    yield* Console.log(table.toString());
+  });
+
+function initTable(): Effect.Effect<Table.Table, never, never> {
+  return Effect.sync(
+    () =>
+      new Table({
```
PetarKirov, line 70: I'd prefer if we could avoid relying on a third-party dependency for this, and instead implement a `drawTable` function in [`@blocksense/base-utils/tty`](https://github.com/blocksense-network/blocksense/blob/main/libs/ts/base-utils/src/tty.ts), similar to [`drawBox`](https://github.com/blocksense-network/blocksense/blob/ae7f8e183c6b1959f4eb00f1112542e79d32a9c9/libs/ts/base-utils/src/tty.ts#L339).

apps/changelog-generator/src/Cli.ts
```diff
@@ -1,8 +1,108 @@
...
 
-const command = Command.make('hello');
+const listFeeds = Command.make(
+  'list-feeds',
+  {
+    dir: Options.directory('dir').pipe(Options.withDefault(configDir)),
+  },
+  ({ dir }) =>
+    Effect.gen(function* (_) {
+      const network: NetworkName = yield* _(getNetwork);
```
PetarKirov, line 21: "Promote" `network` to a CLI option, so that it could be specified via the built-in Effect CLI Wizard.

apps/changelog-generator/src/Cli.ts
```diff
@@ -1,8 +1,108 @@
...
+      const config = yield* _(
```
PetarKirov, line 22: Is wrapping `Effect.tryPromise(..)` in `_(..)` necessary? Can you try if you can do 
`yield* Effect.tryPromise(..)`?

apps/changelog-generator/src/Cli.ts
```diff
@@ -1,8 +1,108 @@
...
+        Effect.tryPromise(() => readConfig('feeds_config_v2', dir)),
+      );
+      yield* _(printFeeds(config, network));
```
PetarKirov, line 25: Nit: I would inline `printFeeds` as it's only called here (i.e. not sure if would be reused in the future in other places). If the entire `Cli.ts` file becomes too long, we can split it in a separate file per sub-command.

apps/changelog-generator/src/Cli.ts
```diff
@@ -1,8 +1,108 @@
...
+  name: 'Changelog Generator',
   version: '0.0.0',
 });
+
+const printFeeds = (
+  config: ConfigType<'feeds_config_v2'>,
+  network: NetworkName,
+) =>
+  Effect.gen(function* (_) {
+    const deploymentData = yield* _(getDeploymentData(network));
```
PetarKirov, line 43: Inline `getDeploymentData` here:
```suggestion
    const deploymentData = yield* Effect.promise(() => readEvmDeployment(network));
```

apps/changelog-generator/src/Cli.ts
```diff
@@ -1,8 +1,108 @@
...
+    if (!deploymentData) {
+      yield* Console.error(
+        `Deployment data for v2 not found for network: '${network}'`,
+      );
+      return;
+    }
```
PetarKirov, line 49: Btw, `readEvmDeployment` has an optional parameter `throwIfNotFound` to throw an error if no data is available for the specified network. I wonder what happens if we omit the error handling here (and pass `true` to `throwIfNotFound`).

apps/changelog-generator/src/Cli.ts
```diff
@@ -1,8 +1,108 @@
...
+
+    const table = yield* _(initTable());
+
+    config.feeds.forEach(feed => {
+      const contract =
+        deploymentData?.contracts?.CLAggregatorAdapter[Number(feed.id)];
```
PetarKirov, line 55:
```suggestion
        deploymentData?.contracts?.CLAggregatorAdapter[`${feed.id}`];
```

A feed id can be larger than [`Number.MAX_SAFE_INTEGER`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) (`2**53 – 1`), as it is defined as `u128`, so a conversion to `number` could be a bug. Also the keys are strings, so you can just call `.toString()` on the id, or better yet - interpolate it.

apps/changelog-generator/src/Cli.ts
```diff
@@ -1,8 +1,108 @@
...
+  Effect.gen(function* (_) {
+    const deploymentData = yield* _(getDeploymentData(network));
+    if (!deploymentData) {
+      yield* Console.error(
+        `Deployment data for v2 not found for network: '${network}'`,
+      );
+      return;
+    }
+
+    const table = yield* _(initTable());
```
PetarKirov, line 51:
```suggestion
    const table = new Table({
      head: [
        'Feed Name',
        'Address',
        'Decimals',
        'Base Address',
        'Quote Address',
      ],
      style: {
        head: [],
      },
      colWidths: [18, 44, 10, 44, 44],
      wordWrap: true,
    });
```

I don't think we need to wrap each normal function call in `Effect.sync`, so we can inline `initTable`, as I also don't think it would be reused.

apps/changelog-generator/src/Cli.ts
```diff
@@ -1,8 +1,108 @@
...
+import { Terminal } from '@effect/platform';
 
-const command = Command.make('hello');
+const listFeeds = Command.make(
+  'list-feeds',
+  {
+    dir: Options.directory('dir').pipe(Options.withDefault(configDir)),
+  },
+  ({ dir }) =>
+    Effect.gen(function* (_) {
```
PetarKirov, line 20: I just read the Effect TS docs about this `_` parameter. TL;DR you don't need it, as we're using TS 5.8+. 
https://effect.website/docs/getting-started/using-generators/#adapter

PetarKirov (commented):
Despite all the comments below, I think you did a fine job with this PR. I'm trying to provide as much feedback as possible, so you can we can all learn together the best practices and incorporate them in our work as soon as possible.

apps/changelog-generator/src/Cli.ts
DannyStoyanov: In my opinion such inlining brings noise in the functions and the whole reading experience is getting worse rather than extracting separate functionalities into functions with **meaningful names**. If the extracted function contains a few number of lines I would be more inclined to inline the function.

For now I will inline it but we can discuss it for future cases.

PetarKirov: > we can discuss it for future cases.

Of course. Inlining vs extracting logic in functions with a meaningful name is a depends on the circumstance and is a judgement call.

MillaGenova (approved):

PR #1444, opened 2025-06-23
[direnv]: Improve `direnv` Developer Experience

EmilIvanichkovv:
### Improve `direnv` Stability and Developer Experience

---

#### 1. **Upgrade `nix-direnv` to v3.1.0**

Brings in the latest improvements and fixes from upstream.

---

#### 2. **Reduce Terminal Noise from `direnv`**

* Sets `DIRENV_LOG_FORMAT=""` to suppress verbose output like `direnv: export +VAR -VAR ...`.
* Prevents scrollback clutter, especially useful in setups that export many variables.

Here’s what `direnv reload` looks like before and after:

| Before                                                                                     | After                                                                                     |
| ------------------------------------------------------------------------------------------ | ----------------------------------------------------------------------------------------- |
| before | after |

---

#### 3. **Exit on Fallback Activation**

* Adds a guard for `NIX_DIRENV_DID_FALLBACK=1`, which signals that `nix-direnv` failed to load the expected environment.
* If this fallback is triggered, we now exit to avoid silently entering a broken or stale shell.
* This makes failures explicit and prevents subtle bugs caused by partially loaded environments (e.g., when `devShell` builds fail).

Example:

| Before                                                                                     | After                                                                                     |
| ------------------------------------------------------------------------------------------ | ----------------------------------------------------------------------------------------- |
| before | after |

---

### Why This Matters

* Without the fallback check, it's easy to end up in a broken shell without noticing — leading to subtle and frustrating bugs.
* Silencing the env var spam helps focus attention on meaningful errors or output. In practice, this means you’ll actually **see** when your shell fails to load instead of losing it in a sea of `export` logs.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -9,3 +9,6 @@ fi
 dotenv_if_exists
 
 use flake ".#${DEV_SHELL:-default}" --impure
+if [[ "$NIX_DIRENV_DID_FALLBACK" = 1 ]]; then
+  exit 1
+fi


DannyStoyanov (approved):
lgtm

MillaGenova (approved):

PetarKirov (approved):

PR #1413, opened 2025-06-11
[changelog-generator]: Init `@blocksense/changelog-generator` workspace

DannyStoyanov:
This new app is created via: `yarn create effect-app@latest` which comes with pre-configured:
- `effect`
- `changesets`
- `eslint`
- `nix flake`
- `github workflow` - effect specific
- `vitest`

We took advantage of `Effect`'s initial setup and right after that we have examined the default settings and modified them to suit our needs and setup.

In the end this app setup will be used as a template for introducing new `Effect` apps within the `blocksense` project.

PR #1449, opened 2025-06-25
feat(avm-relayer): Initialize AVM relayer project.

NikolayKostadinov21:
Followed #1413's changelog-generator project implementation to create the AVM relayer.


apps/avm-relayer/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@blocksense/avm-relayer",
+  "version": "0.0.1",
```
DannyStoyanov, line 3: nit: we can start with `"version": 0.0.0` and each time we introduce new command/subcommand we can increment the MINOR version (`MAJOR.MINOR.PATCH`, reference: https://semver.org/). E.g. when initializing the `changelog-generator` we began with `"version": 0.0.0` and after adding the `list-feeds` command we incremented the version to be `"version": 0.1.0`

apps/avm-relayer/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "type": "module",
+  "license": "MIT",
+  "description": "The AVM Relayer that forwards feed data from Kafka's message queue to the ADFS contract.",
+  "bin": "./dist/bin.cjs",
+  "scripts": {
+    "build": "tsup --dts && yarn copy-package-json",
+    "ts": "yarn node --import tsx",
```
DannyStoyanov, line 10: Here you can change `ts` command to `start` [the way it is merged in main](https://github.com/blocksense-network/blocksense/blob/main/apps/changelog-generator/package.json#L10)
```suggestion
   "start": "yarn node --import tsx ./src/bin.ts",
```

apps/avm-relayer/src/Cli.ts
```diff
@@ -0,0 +1,8 @@
+import * as Command from '@effect/cli/Command';
+
+const command = Command.make('hello relayer');
+
+export const run = Command.run(command, {
+  name: 'AVM Relayer',
+  version: '0.0.1',
```
DannyStoyanov, line 7: nit: again in relation of the comment in `package.json`, we can change the version to `0.0.0` when initializing the app

apps/avm-relayer/src/interfaces.ts
DannyStoyanov, line 1: I would suggest this PR to focus only on initializing the `avm-relayer` and this file to be included in following PR with the functionality which is using these interfaces. 

Apart from that, I think we have agreed to use `type` instead of `interface` unless you have a strong reason to use `interface`, I am mentioning this just to keep it in mind for future PRs.

apps/avm-relayer/src/utils.ts
DannyStoyanov, line 1: Same as previous comment: let's focus only on initializing the `avm-relayer` app and latter on you can add `utils.ts` file.

apps/avm-relayer/tsconfig.json
DannyStoyanov, line 1: I think you should exclude the `dist` folder as we have done it [here](https://github.com/blocksense-network/blocksense/blob/main/apps/changelog-generator/tsconfig.json#L13).

apps/avm-relayer/tsconfig.test.json
DannyStoyanov, line 1: I see differences with the code in main here. In main we have [this setup](https://github.com/blocksense-network/blocksense/blob/main/apps/changelog-generator/tsconfig.test.json). First, you somehow miss `"rootDir": "test",` in the `compilerOptions`. Secondly - I don't understand why do we add `src` in the `include` property. We have separate `tsconfig.src.json` for it and here`src` is redundant in my opinion.

DannyStoyanov (changes requested):
My only concern regards this PR is to focus entirely on the app initialization and every other logic not related with that(e.g. `src/utils.ts` and `src/interfaces.ts`) to be included in following PRs where this functionality is necessary. The remaining comments are minor and intended to maintain a consistent setup across all newly initialized apps.

DannyStoyanov:
You can drop the [fix(avm-relayer): Fix red underlying warnings from vscode.](https://github.com/blocksense-network/blocksense/pull/1449/commits/0a45b410588bd717f2f3ef846a4ad7821a470088) commit as I have already fixed this issue this [this commit](https://github.com/blocksense-network/blocksense/pull/1429/commits/5321cc739dbd06d0d503e7832706dcbbb3bea067) which is already merged in main. So you can rebase with main order to get it.

apps/avm-relayer/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "@effect/experimental": "^0.48.10",
+    "@effect/platform": "0.84.9",
+    "@effect/platform-node": "0.85.14",
+    "@effect/printer": "^0.44.5",
+    "@effect/printer-ansi": "^0.44.5",
+    "@effect/rpc": "^0.61.13",
+    "@effect/sql": "^0.37.10",
+    "@effect/typeclass": "^0.35.5",
+    "effect": "^3.15.1",
+    "kafkajs": "^2.2.4"
```
DannyStoyanov, line 35: nit: remove this dependency for now and introduce it once it is necessary

apps/avm-relayer/src/Cli.ts
```diff
@@ -0,0 +1,8 @@
+import * as Command from '@effect/cli/Command';
+
+const command = Command.make('hello relayer');
+
+export const run = Command.run(command, {
+  name: 'AVM Relayer',
+  version: '0.0.0',
```
DannyStoyanov, line 7: nit: you can reuse `package.json`'s values as implemented [here](https://github.com/blocksense-network/blocksense/blob/main/apps/changelog-generator/src/Cli.ts#L12-L13). You just have to set the `description` property to `AVM Relayer`.

DannyStoyanov (changes requested):
Lastly, please copy `.changesets` and `.vscode` files from `/apps/changelo-generator` or from `apps/chain-interactions` and add them to `apps/avm-relayer`. After that and the small nits, I think you are ready to go! :)

DannyStoyanov (approved):

PR #1457, opened 2025-06-27
[config-types]: Add `dfcg-artifacts-downloader` for reading old configurations

DannyStoyanov:
### Motivation 💡 
This is pre-step before comparing the data of the new deployment  with the old one. As we are going to remove all V1 related configurations and files, we have uploaded them to the [`dfcg-artifacts` repo ](https://github.com/blocksense-network/dfcg-artifacts)in separate `/configs` folder. For that reason we need a way to read those files and here comes the new `dfcg-artifacts-downloader` file which is introduced in the current PR. 

### Features ✨ 
**`config-types`**
- Added `dfcg-artifacts-downloader.ts`:
 - Uses `@octokit/rest` to download config files (`feeds_config_v1`, `sequencer_config_v1`, `evm_contracts_deployment_v1`, `chainlink_compatibility_v1.json`) from the `dfcg-artifacts` GitHub repo.
- Refactored `configFiles`
- Parses and validates downloaded configs using effect schemas.
- Dependencies
  - Installed `@octokit/rest@21.1.1
- Added node-config module:
  - Introduced SequencerConfigV1Schema and its corresponding type.
- Exported DeploymentConfigV1 type from evm-contracts-deployment.

[BSN-3193: Read old v1 configs from dfcg-artifacts repo](https://coda.io/d/_d6vM0kjfQP6#_tuk5j/r3193&view=full)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.repository }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
+env:
+  DFCG_ARTIFACTS_ACCESS_TOKEN: ${{ secrets.DFCG_ARTIFACTS_ACCESS_TOKEN }}
+
 jobs:
   lint:
     runs-on: [self-hosted, nixos, x86-64-v3]
--- a/.github/workflows/on-main.yml
+++ b/.github/workflows/on-main.yml
@@ -5,6 +5,9 @@ on:
     branches:
       - main
 
+env:
+  DFCG_ARTIFACTS_ACCESS_TOKEN: ${{ secrets.DFCG_ARTIFACTS_ACCESS_TOKEN }}
+
 jobs:
   deploy_websites:
     name: Deploy ${{ matrix.project }} website


libs/ts/config-types/src/node-config/types.ts
```diff
@@ -0,0 +1,87 @@
...
+  multicall_contract_address: S.NullishOr(S.String),
+  transaction_retries_count_limit: S.Number,
+  transaction_retry_timeout_secs: S.Number,
+  retry_fee_increment_fraction: S.Number,
+  transaction_gas_limit: S.Number,
+  data_feed_store_byte_code: S.NullishOr(S.String),
+  data_feed_sports_byte_code: S.NullishOr(S.String),
+  impersonated_anvil_account: S.NullishOr(S.String),
+  is_enabled: S.Boolean,
+  allow_feeds: S.NullishOr(S.Array(S.Number)),
```
EmilIvanichkovv, line 55: I think this should be BigInt. Not 100% sure tho

DannyStoyanov: Turned out `S.BigIntFromNumber` is the correct one.

libs/ts/config-types/src/node-config/types.ts
```diff
@@ -0,0 +1,100 @@
+import { Schema as S, pipe } from 'effect';
+
+// This is a partial schema including only allowed feeds
+// TODO: (danielstoyanov) Expand to whole config schema
+export const SequencerConfigV1Schema = S.mutable(
+  S.Struct({
+    providers: S.Record({
+      key: S.String,
+      value: S.Struct({
+        allow_feeds: S.NullishOr(S.Array(S.Number)),
```
PetarKirov, line 10:
```suggestion
        allow_feeds: S.optional(S.Array(S.Number)),
```

libs/ts/base-utils/src/schemas.ts
```diff
@@ -97,3 +97,11 @@ export const fromCommaSeparatedString = <T, S extends string>(
       },
     },
   );
+
+export const PortSchema = S.Int.pipe(
```
EmilIvanichkovv, line 101: nice

libs/ts/config-types/tests/config-decoding.spec.ts
```diff
@@ -11,18 +12,58 @@ import {
...
+
+const LEGACY_CONFIGS_DIR = 'configs';
+
+const tokenValid = await isTokenValid();
+if (!tokenValid) {
+  console.warn(
+    `Skipping tests for legacy configs. Renew GitHub token for /blocksense/dfcg-artifacts repo`,
+  );
+}
+describe.skipIf(!tokenValid)(
```
EmilIvanichkovv, line 40:
```suggestion

describe.skipIf(!tokenValid)(
```

EmilIvanichkovv (approved):
Good work. Left a nit

libs/ts/config-types/tests/config-decoding.spec.ts
```diff
@@ -11,18 +12,58 @@ import {
...
+  isTokenValid,
+} from '../src/dfcg/artifacts/downloader';
+import { join } from 'path';
+import { rootDir } from '../../base-utils/src/env/constants';
+
+const LEGACY_CONFIGS_DIR = 'configs';
+
+const tokenValid = await isTokenValid();
+if (!tokenValid) {
+  console.warn(
```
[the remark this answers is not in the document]
EmilIvanichkovv: Yes, yes, there is no problem. What I meant is that choosing the alternative approach, which omits the token validity check, would block the CI.

EmilIvanichkovv (approved):

libs/ts/config-types/src/node-config/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from '../dfcg/artifacts/downloader';
```
PetarKirov, line 1:
```suggestion
```

libs/ts/config-types/src/dfcg/artifacts/downloader.ts
```diff
@@ -0,0 +1,108 @@
...
+  return {
+    sequencerDeploymentConfigV1: {
+      providers: sequencerData.providers,
+    },
+    feedsConfigV1: {
+      feeds: feedsData.feeds,
+    },
+    evmContractsDeploymentV1: evmContractsData,
+    chainlinkDeploymentV1: chainlinkData,
+  };
```
PetarKirov, line 89: You can use `mapValuePromises` from #1521:

```suggestion
  const legacyConfigFiles = {
    sequencerDeploymentConfigV1: 'sequencer_config_v1',
    feedsConfigV1: 'feeds_config_v1',
    evmContractsDeploymentV1: 'evm_contracts_deployment_v1',
    chainlinkDeploymentV1: 'chainlink_compatibility_v1',
  } satisfies Record<string, ConfigFileName>;
  
  return mapValuePromises(obj, (key, cfgFileName) => 
    fetchConfigFromDfcgRepo(cfgFileName)
  );
```

libs/ts/config-types/tests/config-decoding.spec.ts
```diff
@@ -11,18 +12,59 @@ import {
...
+    entriesOf(configTypes).filter(([key]) => !(key in legacyConfigTypes)),
+  ) satisfies Record<string, S.Schema<any>>;
+
+  for (const configType of keysOf(configTypesWithoutLegacy)) {
+    it(`should decode '${configType}' config file successfully`, async () => {
+      const dir =
+        configType === 'sequencer_config_v2'
+          ? join(rootDir, '/libs/ts/config-types/tests/fixtures')
+          : configDirs[configType];
+      const config = await readConfig(configType as ConfigFileName, dir);
```
PetarKirov, line 67: By using `keyOf` from `base-utils`, TypeScript should be able to infer the type key type of `configTypes`, so that you wouldn't need to cast it.
Also using an "early return" (`continue` in this case) is cleaner:
 
```suggestion
  for (const configType of keysOf(configTypes)) {
    if (configType in legacyConfigTypes) continue;
    it(`should decode '${configType}' config file successfully`, async () => {
      const dir =
        configType === 'sequencer_config_v2'
          ? join(rootDir, '/libs/ts/config-types/tests/fixtures')
          : configDirs[configType];
      const config = await readConfig(configType, dir);
```

PetarKirov (commented):
A couple of more opportunities to improve the code:

PR #1465, opened 2025-07-01
feat(sequencer): Add round counters per provider

yordanmadzhunkov:
feat(sequencer): Add round counters per provider

* Reading round counters from adfs contract and latest published values
* Add test that reads round values from chain
* Refactor how smart contracts are configured in sequencer config
* Fix(sequencer/config): Configuration related to contracts for second round consensus made consistent
* Fix example for process compose
* Use function pointers and chunked calls to multicall contract
* Fix changes that were causing a mismatch between sequencer and reporter when generating calldata
* Fix(e2e-tests): Fix failing tests due to changes in the config and process compose setup
* Add config option for reading values and round counteres
* feat(sequencer/providers): Read data from networks in parallel

To speed up initialization in the sequencer we read data in parallel from
all networks. We also change the readiness_probe to not have a 120
second initial delay, but probe every 10 seconds and require 10
consecutive failing probes in order to restart the process.

Co-authored-by: Hristo Staykov 
Co-authored-by: Aneta Tsvetkova 

Corresponding [PR-524 in Infra](https://github.com/blocksense-network/infra/pull/524) repo


HristoStaykov:
Re-based on top of **main** to resolve conflicts related to merging [PR-1486](https://github.com/blocksense-network/blocksense/pull/1486)

apps/sequencer/sequencer_config.json
```diff
@@ -13,28 +13,66 @@
     "ETH1": {
       "private_key_path": "/tmp/priv_key_test",
       "url": "http://127.0.0.1:8545",
-      "transaction_retries_count_limit": 5,
```
HristoStaykov, line 16: Nit: At some point during PR reviews and discussions I think we decided to use `transaction_retries_count_limit` instead of `transaction_retries_count_before_give_up`. I think we might want to be consistent.

apps/sequencer/src/feeds/votes_result_sender.rs
```diff
@@ -193,7 +195,21 @@ async fn try_send_aggregation_consensus_trigger_to_reporters(
...
                 info!("Network `{net}` is enabled; initiating second round consensus");
             }
-
+            if provider_settings
+                .get_contract_config(GNOSIS_SAFE_CONTRACT_NAME)
+                .is_none()
+            {
+                info!("Network `{net}` not configured for second round consensus - skipping");
+                continue;
+            }
```
HristoStaykov, line 204: I see that we have above (Line 182) another check `if provider_settings.safe_address.is_none()` that we might want to reconsider and potentially remove.

libs/ts/config-types/src/node-config/types.ts
```diff
@@ -39,15 +39,22 @@ const PublishingCriteriaSchema = S.Struct({
   peg_tolerance_percentage: S.Number,
 });
 
+const ContractSchema = S.Struct({
+  name: S.String,
+  address: S.NullishOr(S.String),
```
EmilIvanichkovv, line 44: In what case we have contract name but no address?

libs/ts/config-types/src/node-config/types.ts
```diff
@@ -39,15 +39,22 @@ const PublishingCriteriaSchema = S.Struct({
   peg_tolerance_percentage: S.Number,
 });
 
+const ContractSchema = S.Struct({
+  name: S.String,
```
EmilIvanichkovv, line 43: Do we know the names of the contracts that can be passed here. If so lets make it string literal

libs/ts/config-types/src/node-config/types.ts
```diff
@@ -39,15 +39,22 @@ const PublishingCriteriaSchema = S.Struct({
...
+  address: S.NullishOr(S.String),
+  creation_byte_code: S.NullishOr(S.String),
+  deployed_byte_code: S.NullishOr(S.String),
+  min_quorum: S.NullishOr(S.Number),
+  contract_version: S.Number,
```
EmilIvanichkovv, line 48: Should we keep the support of the old version of our contracts?

libs/ts/config-types/src/node-config/types.ts
```diff
@@ -39,15 +39,22 @@ const PublishingCriteriaSchema = S.Struct({
...
   private_key_path: S.String,
   url: S.String,
-  contract_address: S.NullishOr(S.String),
-  safe_address: S.NullishOr(S.String),
-  safe_min_quorum: S.Number,
   event_contract_address: S.NullishOr(S.String),
   multicall_contract_address: S.NullishOr(S.String),
-  transaction_retries_count_limit: S.Number,
+  transaction_retries_count_limit: S.NullishOr(S.Number),
+  transaction_retries_count_before_give_up: S.NullishOr(S.Number),
```
EmilIvanichkovv, line 57: These can not be nullish I think.

https://github.com/blocksense-network/blocksense/blob/714c668e6f3bbcc5af4d809ec1cf96e374c7d6c3/apps/sequencer/src/providers/provider.rs#L82-L83

nix/modules/blocksense/module-opts/sequencer/provider/contract.nix
```diff
@@ -0,0 +1,39 @@
...
+      type = types.str;
+      default = "";
+      description = mdDoc "Name of the contract";
+    };
+
+    address = mkOption {
+      type = types.nullOr types.str;
+      default = null;
+      description = mdDoc "Address of the contract";
+    };
```
EmilIvanichkovv, line 13: ditto. same question

nix/shells/pkg-sets/rust.nix
```diff
@@ -12,7 +11,6 @@
     REPORTER_CONFIG_DIR = config.devenv.root + "/apps/reporter";
     FEEDS_CONFIG_DIR = config.devenv.root + "/config";
     REPORTER_SECRET_KEY_FILE_PATH = config.devenv.root + "/nix/test-environments/test-keys";
-    LD_LIBRARY_PATH = lib.makeLibraryPath self'.legacyPackages.commonLibDeps;
```
EmilIvanichkovv, line 15: You remove it from here but add in `nix/modules/blocksense/backends/process-compose.nix`

Why is this needed?

nix/test-environments/example-setup-01.nix
```diff
@@ -113,11 +112,41 @@ in
...
+              creation-byte-code = null;
+              deployed-byte-code = null;
+              contract-version = 1;
+            }
+            {
+              name = "multicall";
+              address = "0xcA11bde05977b3631167028862bE2a173976CA11";
+              creation-byte-code = null;
+              deployed-byte-code = null;
+              contract-version = 3;
```
EmilIvanichkovv, line 128: What is version 3?

I kinda have idea for 1 and 2. Is 3 related to contracts that are not deployed/developed by us or?

nix/test-environments/example-setup-03.nix
```diff
@@ -73,6 +72,22 @@ in
...
+              contract-version = 2;
+            }
+            {
+              name = "multicall";
+              address = "0xcA11bde05977b3631167028862bE2a173976CA11";
+              creation-byte-code = null;
+              deployed-byte-code = null;
+              contract-version = 3;
+            }
+          ];
```
EmilIvanichkovv, line 90: This seems fine

EmilIvanichkovv: Yet `multicall` probably is different for each network. Here we fork `ink-sepolia` but in the future we might change that. We might need to read if from a file in the future.

nix/test-environments/example-setup-03.nix
```diff
@@ -5,8 +5,6 @@
 }:
 let
   # Function to read and parse the JSON file
-  readJson = path: builtins.fromJSON (builtins.readFile path);
-
```
EmilIvanichkovv, line 9: You moved this below but left the comment. This is util func, so lets keep it on the top.

Maybe don't change the order of the vars below too.

EmilIvanichkovv (changes requested):
First iteration of my review. Left some comments.

nix/modules/blocksense/backends/process-compose.nix
```diff
@@ -111,14 +112,15 @@ let
           curl -fsSL http://127.0.0.1:${toString cfg.sequencer.ports.admin}/health \
             -H 'content-type: application/json'
         '';
-        initial_delay_seconds = 0;
-        period_seconds = 1;
+        initial_delay_seconds = 120;
+        period_seconds = 10;
```
EmilIvanichkovv, line 116: Need more insight on this change.

libs/ts/config-types/src/node-config/types.ts
yordanmadzhunkov: Contract exists, but it's not deployed. When the sequencers deploys the contract, it sets the address

EmilIvanichkovv: I think it never deploys. We do all deployments with the hardhat script, no?

libs/ts/config-types/src/node-config/types.ts
yordanmadzhunkov: Maybe, they can be removed in some other PR. I don't want to do it here, because it will take more work to rewrite all existing test to use new contracts

libs/ts/config-types/src/node-config/types.ts
yordanmadzhunkov: I agree, only one of them should be provided. I am not sure how to specify that in `types.ts`

nix/modules/blocksense/module-opts/sequencer/provider/contract.nix
yordanmadzhunkov: same answer :D

nix/test-environments/example-setup-01.nix
yordanmadzhunkov: 3 comes from their documentation ( and obviously, some day we may have logic based on this number). Currently we use `agreggate` contract call which can be also found in previous version of `multicall`. 

[Multicall3](https://github.com/mds1/multicall3/blob/main/src/Multicall3.sol) is deployed on over 250 chains at 0xcA11bde05977b3631167028862bE2a173976CA11. The full list of deployed chains along with the Multicall3 ABI can be found at https://multicall3.com/. The ABI is provided in various formats, and can be copied to your clipboard or downloaded to a file.

nix/test-environments/example-setup-03.nix
yordanmadzhunkov: ok

nix/shells/pkg-sets/rust.nix
yordanmadzhunkov: because this caused problems with dev-shell:
```
8.13.0/lib/libcurl.so.4: undefined symbol: libssh2_session_callback_set2
```

EmilIvanichkovv (commented):
Tested with `just start-environment example-setup-01` and can confirm the rounds are being successfully read!

nix/test-environments/example-setup-03.nix
```diff
@@ -72,6 +70,22 @@ in
...
+              contract-version = 2;
+            }
+            {
+              name = "multicall";
+              address = "0xcA11bde05977b3631167028862bE2a173976CA11";
+              creation-byte-code = null;
+              deployed-byte-code = null;
+              contract-version = 3;
+            }
+          ];
```
EmilIvanichkovv, line 88: When I remove the `multicall` contract from the config, the sequencer still manages to parse the rounds. How is that possible?

EmilIvanichkovv: Ahaa.. `load_round_counters_from_chain` does not use multicall

EmilIvanichkovv: So, what do we use `multicall` for?

libs/config/src/lib.rs
```diff
@@ -170,37 +177,49 @@ impl PublishCriteria {
...
     /// Whether data is written to the network (provider) or not. Useful for devops, if a single
     /// network is acting up and needs to be disabled without restarting the whole service.
     #[serde(default = "default_is_enabled")]
     pub is_enabled: bool,
 
+    #[serde(default = "default_is_enabled")]
+    pub should_load_round_counters: bool,
+
+    #[serde(default = "default_is_enabled")]
+    pub should_load_historical_values: bool,
```
EmilIvanichkovv, line 213: Can we add as nix options?

EmilIvanichkovv:
> Testing the update of the rounds came with weird behaviour image
> 
> Need to investigate more

Tested more. The logic around counters seems fine.
Soon will add checks in the e2e tests.

EmilIvanichkovv (commented):
I think we should update the confis in the `infra`
cc: @PetarKirov @MartinNikov

nix/test-environments/example-setup-03.nix
yordanmadzhunkov: Initial implementation both round counters and history values used multicall. Then I switched the reading of round counters with faster ( in theory ) call `get_storage_at`. Maybe the same is possible for reading the historical values in the ADFS contract. In the legacy contract, we still use mullticall for reading historical values.

libs/config/src/lib.rs
```diff
@@ -595,28 +713,65 @@ mod tests {
...
+                    "contract_version": 1
+                }
             ]
             }"#).unwrap();
         assert!(p.is_enabled);
-        assert_eq!(p.event_contract_address, None);
+        // assert_eq!(p.event_contract_address, None);
```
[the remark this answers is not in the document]
yordanmadzhunkov: done

libs/config/src/lib.rs
```diff
@@ -170,37 +177,49 @@ impl PublishCriteria {
...
     pub private_key_path: String,
     pub url: String,
-    pub contract_address: Option<String>,
-    pub safe_address: Option<String>,
-    pub safe_min_quorum: u32,
-    pub event_contract_address: Option<String>,
-    pub multicall_contract_address: Option<String>,
-    pub transaction_retries_count_limit: u32,
+    #[serde(alias = "transaction_retries_count_limit")]
+    pub transaction_retries_count_before_give_up: u32,
```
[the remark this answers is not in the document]
yordanmadzhunkov: removed

libs/config/src/lib.rs
yordanmadzhunkov: done

EmilIvanichkovv (commented):
While extending the e2e tests with sequencer round parsing logic, I noticed a weird behaviour:
Sequencer incorrectly overwrites the initial round (e.g., if the latest round (read from the network) is 42, it overwrites it instead of proceeding with a new round 43). 
More info you can find in my PR https://github.com/blocksense-network/blocksense/pull/1554

However I think we can fix this in follow up PR

apps/sequencer/src/providers/eth_send_utils.rs
```diff
@@ -350,7 +348,7 @@ pub async fn eth_batch_send_to_contract(
     feed_type: Repeatability,
     feeds_config: Arc<RwLock<HashMap<FeedId, FeedConfig>>>,
     transaction_retry_timeout_secs: u64,
-    transaction_retries_count_limit: u64,
+    transaction_retries_count_before_give_up: u64,
```
Owliie, line 351: Isn't simply `transaction_retries_count` enough?

apps/sequencer/src/providers/eth_send_utils.rs
```diff
@@ -1175,7 +1184,7 @@ async fn log_round_counters(
     let mut debug_string =
         format!("{prefix} for net = {net} and updated_feeds = {updated_feeds:?} ");
     for feed in updated_feeds {
-        let round_index = round_counters.get(feed).unwrap_or(&0_u64);
+        let round_index = round_counters.get(feed).unwrap_or(&0);
```
Owliie, line 1187: `round` was considered inappropriate for the ring buffer index by @zah. Use `index` or ring buffer index when referring to `round`.

Owliie (dismissed):

EmilIvanichkovv (approved):
Lets ship it

PR #1475, opened 2025-07-03
[oracles/data providers sdk]: Migrate fetchers

MillaGenova:
### Refactor: Centralize Price Fetching Logic in `data_providers_sdk`

This PR moves and unifies the price fetcher logic for both **CEX** and **stock market** oracles into the shared `data_providers_sdk` lib.
It simplifies price collection logic across apps and prepares for future reuse.

---

### What’s Changed

* Moved all CEX and stock fetcher modules to `libs/data_providers_sdk/src/price_data/fetchers/`
* Introduced a shared `fetch_all_prices()` utility to collect and log results from all providers
* Updated `cex-price-feeds` and `stock-price-feeds` apps to use the new SDK fetchers
* Cleaned up duplicated logic and reduced per-oracle boilerplate
* Added missing dependencies (`serde-this-or-that`, `itertools`, `blocksense-sdk`)

[BSN-3234: Move all cex-price-feeds oracle data providers fetchers](https://coda.io/d/_d6vM0kjfQP6#_tuk5j/r3234&view=full)

[BSN-3236: Move all stock-price-feeds oracle data providers fetchers](https://coda.io/d/_d6vM0kjfQP6#_tuk5j/r3236&view=full)


libs/data_providers_sdk/src/price_data/fetchers/stock_markets/utils.rs
```diff
@@ -0,0 +1,18 @@
...
+        "[{network}] Skipping symbol {}: missing {}{}{}",
+        symbol,
+        if price.is_none() { "price" } else { "" },
+        if price.is_none() && volume.is_none() {
+            " and "
+        } else {
+            ""
+        },
+        if volume.is_none() { "volume" } else { "" },
+    );
```
reo101, line 17: Something like this may be better for clarity

```suggestion
    let mut missing_fields = Vec::new();

    if price.is_none() {
        missing_fields.push("price");
    }
    if volume.is_none() {
        missing_fields.push("volume");
    }

	// Do we actually want this (the printing) to happen all the time?
    if !missing_fields.is_empty() {
        eprintln!(
            "[{network}] Skipping symbol {symbol}: missing {}",
            missing_fields.join(" and ")
        );
    }
```

Is this function called only when there **is** an error for sure (i.e. there's always at least one `None` in `price` and `volume`)?

libs/data_providers_sdk/src/price_data/fetchers/fetch.rs
```diff
@@ -0,0 +1,35 @@
...
+use futures::Stream;
+
+use crate::price_data::{traits::prices_fetcher::PairPriceData, types::ProviderPriceData};
+
+pub async fn fetch_all_prices<S>(mut futures_set: S) -> Result<Vec<ProviderPriceData>>
+where
+    S: Stream<Item = (&'static str, Result<PairPriceData>)> + Unpin,
+{
+    let before_fetch = Instant::now();
+    let mut all_fetched_prices: Vec<ProviderPriceData> = Vec::new();
```
reo101, line 14: nit: Maybe don't take the `Vec` allocation time (however small it may be) into account

```suggestion
    let mut all_fetched_prices: Vec<ProviderPriceData> = Vec::new();
    let before_fetch = Instant::now();
```

libs/data_providers_sdk/src/price_data/fetchers/fetch.rs
```diff
@@ -0,0 +1,35 @@
...
+            Ok(prices) => {
+                let time_taken = before_fetch.elapsed();
+                println!("ℹ️  Successfully fetched prices from {exchange_id} in {time_taken:?}",);
+                let prices_per_exchange = ProviderPriceData {
+                    name: exchange_id.to_owned(),
+                    data: prices,
+                };
+                all_fetched_prices.push(prices_per_exchange);
+            }
+            Err(err) => println!("❌ Error fetching prices from {exchange_id}: {err:?}"),
```
reo101, line 28: We never return `Err` from this function.
Are we not supposed to error out on the first unsuccessful fetch?
Or maybe we should collect all the errors and return them instead of/alongside the successfully fetched prices?
If not, we could consider strengthening the return type to just `Vec<ProviderPriceData>`.

reo101: Also, general question, are plain `stdout` writes considered sufficient for error logging?

reo101 (approved):

libs/data_providers_sdk/src/price_data/fetchers/stock_markets/utils.rs
MillaGenova: Any of them or both can be missing

libs/data_providers_sdk/src/price_data/fetchers/fetch.rs
MillaGenova: Let's just go with changing the return type. The main goal is to ensure that if any fetcher fails, it's logged in the oracle, but we don't need to handle the error beyond that. There might be different reasons of fails (e.g. hitting rate limit, slow connection, etc.), but we want to collect data from as many as possible providers. We don't want to return the errors.

As the oracle is being executed in `wasm` currently we don't support tracing and thats why we use `stdout`.

libs/data_providers_sdk/src/price_data/fetchers/fetch.rs
MillaGenova: Resolved this one

libs/data_providers_sdk/src/price_data/fetchers/stock_markets/utils.rs
MillaGenova: With both of them missing we should see `Skipping symbol X: price and volume`

PR #1497, opened 2025-07-10
[docs-ui]: Use `#1f1f1f` as new bg-color for dark theme

RadoslavDimchev:
Before:
image

After:
image

### DIFF
--- a/libs/ts/docs-ui/blocksense-theme/variables.css
+++ b/libs/ts/docs-ui/blocksense-theme/variables.css
@@ -17,7 +17,7 @@
   --breakpoint-3xl: 1920px;
 
   --bg-color: #fffbfa;
-  --bg-color-dark: #2b2929;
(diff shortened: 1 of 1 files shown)


EmilIvanichkovv (approved):

RadoslavDimchev:
> Just a note here(not a request), with `scss` we can take benefit of `variables` and construct `design tokens`/`semantic tokens` which can be reused throughout the front-end websites. I don't know the current progress with the `design system` but with variables we can easily achieve reusability and unified UI experience. They can be used for various things like: colors, different offsets, border styles, shadows and many more.
> 
> Have you thought about something similar for the websites in the blocksense merchandise? Especially for the new ones?

Thanks for the note 😊

We’re using CSS custom properties (variables) too. This setup gives us runtime theming, effortless overrides, and consistent styling, without needing additional tools, but we can think about the benefits of `scss`.

PR #1500, opened 2025-07-10
[.gitignore]: Make sure all `nix/test-environments/test-keys` are ignored

MillaGenova:
And whitelist the ones that we need commited


EmilIvanichkovv:
Another troublemaker off the list

PR #1515, opened 2025-07-17
[contracts-hardhat]: Use apis from networks.ts when verifying contracts

kkirkov:
Now uses the explorer info we have in `networks.ts` when running verify
Where we have more then 1 explorer we can switch to verify 2nd or 3rd with `explorerIndex` : use --explorer-index
Added a new --starting-contract option for when we want to skip verifying some of the adapters

### DIFF
--- a/libs/ts/contracts/tasks/utils.ts
+++ b/libs/ts/contracts/tasks/utils.ts
@@ -87,12 +87,13 @@ export const getApiKeys = () =>
     ]),
   );
 
+// the 'condition' could be used to make the binarySearch - lowerBound, upperBound or custom
 export const binarySearch = <T>(
   array: T[],
-  left: number,
-  right: number,
   condition: (mid: T, index: number) => boolean,
 ): number => {
+  let left = 0;
+  let right = array.length;
   while (left < right) {
     const midIndex = Math.floor((left + right) / 2);
     const midValue = array[midIndex];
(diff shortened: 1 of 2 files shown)


libs/ts/contracts/tasks/utils.ts
```diff
@@ -55,3 +60,45 @@ export const adjustVInSignature = async (
...
+        chainId: meta.chainId,
+        urls: {
+          apiURL: explorer.apiUrl!,
+          browserURL: explorer.webUrl,
+        },
+      };
+    });
+
+export const getApiKeys = () =>
+  Object.fromEntries(
```
DannyStoyanov, line 80: nit: You can make use of [fromEntries](https://github.com/blocksense-network/blocksense/blob/main/libs/ts/base-utils/src/array-iter.ts#L47) method which is already defined in `libs/ts/base-utils/src/array-iter.ts` as it is typed version of `Object.fromEntries`.

libs/ts/contracts/tasks/utils.ts
```diff
@@ -55,3 +60,45 @@ export const adjustVInSignature = async (
...
+    Object.keys(networkMetadata).map(name => [
```
DannyStoyanov, line 81: nit: Same here. Check out [keysOf](https://github.com/blocksense-network/blocksense/blob/main/libs/ts/base-utils/src/array-iter.ts#L10) in `libs/ts/base-utils/src/array-iter.ts`

kkirkov: Thanks, added in a new commit

Owliie (dismissed):

Owliie (approved):

PR #1516, opened 2025-07-17
fix(spin-plugin): calculate `sha256` hashes at build-time

reo101:
This fixes a reproducibility oversight caused by us running a
`builtins.hashFile` on the output of the `spin` plugin derivation which,
by being a Rust derivation, is not (currently) reproducible.

We still do the `tar`-ing in separate derivations thus we do packages
iteration in `Nix`-land (with a big `lib.concatStringsSep` to combine the
resulting scripts), but we do the hashing (now a `sha256sum` call,
instead of a `builtins.hashFile "sha256"`) in `Bash`-land, which removes
our previous erroneous content-addressation of the `spin` plugin.

We now use `jq` to reconstruct the `packages` `JSON` array by starting
out with a `JSON`-ified `manifest` with an empty `packages` array and
slowly doing `jq '.packages += [{ ... }]'` calls over it, thus giving us
a place to do build-time calculation of the `spin` plugins' tarballs'
`sha256` hashes.

### DIFF
--- a/nix/pkgs/spin-plugin/default.nix
+++ b/nix/pkgs/spin-plugin/default.nix
@@ -1,9 +1,17 @@
 {
+  # Lib
+  lib,
+
+  # Helpers
   hostPlatform,
   runCommand,
-  spin,
   writers,
-  lib,
+
+  # Packages
+  spin,
+  jq,
+  moreutils, # sponge
+
   ...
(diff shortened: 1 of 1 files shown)


PetarKirov (approved):
Nice 🔥

PR #1530, opened 2025-07-23
feat(contracs/tasks/ms-change-sequencer): Remove dependency on SafeApiKit

PetarKirov:
This PR introduces two key improvements to the `ms-change-sequencer` hardhat task: it removes the dependency on `SafeApiKit` for broader network compatibility and enhances the user experience with better interactive feedback.

### Changes

-   **feat(contracts/tasks/ms-change-sequencer): Remove dependency on SafeApiKit**
    -   The task's reliance on the Safe transaction service (`SafeApiKit`) has been removed. This change addresses failures on networks that lack full Safe infrastructure support, such as the Taraxa testnet.
    -   Transaction execution is now handled by the `executeMultisigTransaction` utility, making the task more robust and independent of external services.

-   **feat(contracts/tasks/ms-change-sequencer): Improve logging UX**
    -   An interactive confirmation prompt has been added to prevent accidental execution.
    -   Before running, the task now displays a detailed summary box containing:
        -   Admin Multisig configuration (address, threshold, and a list of signers).
        -   The sequencer address being modified.
        -   Whether the sequencer role is being granted or revoked.


libs/ts/contracts/tasks/deployment-utils/access-control.ts
```diff
@@ -201,32 +201,38 @@ export async function setUpAccessControl({
...
-          )
-          .slice(2),
-      operation: OperationType.Call,
-    },
+    ...(!config.reporterMultisig.owners.includes(deployerAddress)
+      ? [
+          {
+            to: reporterMultisigAddress,
+            value: '0',
+            // removeOwner(address prevOwner, address owner, uint256 threshold);
```
EmilIvanichkovv, line 219: is this intentional?

libs/ts/contracts/tasks/deployment-utils/send-tx.ts
```diff
@@ -0,0 +1,59 @@
...
+    };
+  }
+
+  const txRequest: TransactionRequest = {
+    chainId: config.network.chainId,
+    nonce: await config.provider.getTransactionCount(config.deployerAddress),
+    to,
+    value,
+    data,
+    ...(typeof feeData.maxFeePerGas !== 'bigint'
```
EmilIvanichkovv, line 38: Can't we check if null?

EmilIvanichkovv (approved):

libs/ts/contracts/tasks/deployment-utils/access-control.ts
Owliie: Yes, the deployer should not be part of the reporter multisig, so this function removes it

libs/ts/contracts/tasks/deployment-utils/init-chain.ts
```diff
@@ -26,6 +28,7 @@ import type { NetworkConfig } from '../types';
 const sharedPerNetworkKind = {
   deployerAddressIsLedger: asVarSchema(S.BooleanFromString),
   deployerAddress: ethereumAddress,
+  deployerHDWalletDerivationPath: S.String,
```
Owliie, line 31: This should be obligatory if `deployerAddressIsLedger` is true

Owliie (changes requested):

libs/ts/contracts/tasks/deployment-utils/send-tx.ts
```diff
@@ -0,0 +1,59 @@
...
+          type: 0,
+          gasPrice: feeData.gasPrice,
+        }
+      : {
+          maxFeePerGas: feeData.maxFeePerGas,
+          maxPriorityFeePerGas: feeData.maxPriorityFeePerGas,
+        }),
+  };
+
+  const gasLimit = await config.provider.estimateGas(txRequest);
```
Owliie, line 49: It was discussed offline that we should add gas limit as optional env var. I think it would be best to add it here instead of in a new PR

libs/ts/contracts/tasks/deployment-utils/init-chain.ts
Owliie: This should also not be obligatory if `deployerAddressIsLedger` is false, which currently is not the case

Owliie (dismissed):

PR #1540, opened 2025-07-28
[docs.ws/changelog-generator]: Exclude `somnia-mainnet`

RadoslavDimchev:


apps/changelog-generator/src/commands/list-feeds.ts
```diff
@@ -9,7 +9,7 @@ import {
 import { Command, Options } from '@effect/cli';
 import { Effect } from 'effect';
 
-const availableNetworks = await listEvmNetworks();
```
EmilIvanichkovv, line 12: good catch

EmilIvanichkovv (approved):

PR #1543, opened 2025-07-29
[nix): Replace usages of `devenv.root` with `$GIT_ROOT`

MartinNikov:
* Replaced all usages of `config.devenv.root` with `$GIT_ROOT` in Nix configurations, aligning path resolution with Git root discovery at shell startup.
* Make sure `enterShell` scripts is ran before everyting else, asuring `GIT_ROOT` is set before using it.
* Reorganized test key handling:
  * Removed `apps/e2e-tests/test-keys/*` files.
  * At runtime, test keys are now copied from `$GIT_ROOT/nix/test-environments/test-keys` into the working directory of the reporter.
  * Commit API key files with placeholder values
* Updated reporter configuration paths to use `${apiKeysDir}` instead of hardcoded test key paths.
* Changed `logsDir` option type from `path` to `str` to support dynamic Git-root-based paths.


nix/modules/blocksense/backends/process-compose.nix
```diff
@@ -105,7 +102,14 @@ let
...
   sequencerInstance = {
     blocksense-sequencer.process-compose = {
-      command = mkCargoTargetExePath "sequencer";
+      command =
+        ''
+          if [ -z "$FEEDS_CONFIG_DIR" ]; then
+            FEEDS_CONFIG_DIR=${../../../../config}
+          fi
+        ''
+        + mkCargoTargetExePath "sequencer";
```
PetarKirov, line 111:
```suggestion
          fi
          ${mkCargoTargetExePath "sequencer"}
        '';
```

PetarKirov (changes requested):

DannyStoyanov (approved):

PR #1546, opened 2025-07-30
[docs.ws]: Implement email form for future node

RadoslavDimchev:
`FutureNodeForm` has been implemented on the Reporter Node page, enabling users to submit their email addresses via the /newsletter/register endpoint in Cloudflare Worker. These submissions are persisted in the newsletter table backed by a Cloudflare D1 db.

image

### DIFF
--- a/apps/docs.blocksense.network/components/FutureNodeForm.tsx
+++ b/apps/docs.blocksense.network/components/FutureNodeForm.tsx
@@ -1,0 +1,109 @@
+'use client';
+
+import { FormEvent, useCallback, useState } from 'react';
+import { Effect } from 'effect';
+import * as S from 'effect/Schema';
+
+import { EmailSchema } from '@blocksense/base-utils/schemas';
+
+import { getApiClient } from '@/service/client';
+
+export default function FutureNodeForm() {
+  const [email, setEmail] = useState('');
+  const [loading, setLoading] = useState(false);
+  const [success, setSuccess] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+
+  const onRegister = useCallback(
+    (e: FormEvent<HTMLFormElement>) => {
+      e.preventDefault();
+
+      setError(null);
+
+      if (!email || loading || success) return;
+      if (!S.is(EmailSchema)(email)) return;
+
+      setLoading(true);
+      return Effect.runPromise(
+        Effect.gen(function* () {
+          const client = yield* getApiClient();
+          yield* client.newsletter.register({
+            payload: {
+              email,
+              interests: ['node'],
+            },
+          });
+          setSuccess(true);
+        }).pipe(
+          Effect.catchAll(error => {
+            console.error('Registration failed:', error.toString());
+            setError('Registration failed');
+            return Effect.succeed(undefined);
+          }),
+          Effect.tap(() => {
+            setLoading(false);
+            setEmail('');
+          }),
+        ),
+      );
+    },
+    [email, loading],
+  );
+
+  return (
+    <form
+      onSubmit={onRegister}
+      className="p-7 bg-[#e8e8e8] dark:bg-[#484848] rounded-2xl max-w-[50rem] mt-10 mx-auto"
+    >
+      <section className="mb-5">
+        <p className="text-2xl mb-3">
+          {success
+            ? 'Thanks for your interest in running a node with Blocksense!'
+            : 'Sign up as a future node operator'}
+        </p>
+        <p className="max-w-[29.313rem] leading-[22px]">
+          {success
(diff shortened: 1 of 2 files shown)


apps/social-verification/src/newsletter-db-schema.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE
```
DannyStoyanov, line 1: nit: Whenever you are creating a new table it is preferable to use the following syntax: `CREATE TABLE IF NOT EXISTS ...`

apps/social-verification/worker-configuration.d.ts
```diff
@@ -1,14 +1,20 @@
...
     DISCORD_BOT_TOKEN: string;
+    NFT_CLIENT_ID: string;
+    NFT_PRIVATE_KEY: string;
+    NFT_SMART_CONTRACT_ADDRESS: string;
     SOCIAL_DATA_API_KEY: string;
     X_BLOCKSENSE_ACCOUNT_ID: string;
     X_BLOCKSENSE_TWEET_ID: string;
+    SENDGRID_API_KEY: string;
+    DB: D1Database;
+    newsletterDB: D1Database;
```
DannyStoyanov, line 17: nit: shouldn't this be in upper snake case too?

apps/docs.blocksense.network/components/FutureNodeForm.tsx
```diff
@@ -0,0 +1,109 @@
+'use client';
+
+import { FormEvent, useCallback, useState } from 'react';
+
```
MillaGenova, line 4:
```suggestion
```

apps/docs.blocksense.network/service/client.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  const rawClientEffect = HttpApiClient.make(verifyApi, {
+    baseUrl,
+    transformClient: client =>
+      client.pipe(
+        HttpClient.mapRequest(HttpClientRequest.setHeader('x-api-key', apiKey)),
+      ),
+  });
+
+  return rawClientEffect.pipe(Effect.provide(FetchHttpClient.layer));
```
DannyStoyanov, line 32: Provide `FetchHttpClient.layer` whenever you call `getApiClient()` and not here because you might want to use different fetch function for that purpose. In the current scenario you are saying that we need `HttpClient` and you are providing one and in that way you are eliminating the possibility to use different than `FetchHttpClient` layer which breaks the Effect-TS' Dependency Injection mechanism (DI).

I have made the same mistake for `getMetrics` function which I am going to fix in the near future.

In summary, we need to make the function to have 3rd parameter `HttpClient` and whoever uses it, have to provide such client(we use `FetchHttpClient` as Effect provides it and we don't have to write ourselves - only if we need custom we have to provide different one).

apps/docs.blocksense.network/service/client.ts
```diff
@@ -0,0 +1,33 @@
...
+  FetchHttpClient,
+  HttpApiClient,
+  HttpClient,
+  HttpClientRequest,
+} from '@effect/platform';
+
+import { assertNotNull } from '@blocksense/base-utils/assert';
+import { verifyApi } from '@blocksense/social-verification/api';
+
+export function getApiClient() {
```
DannyStoyanov, line 14: Create the whole function to be an Effect using the `Effect.gen()` generator. In that way you can specify the type of the Effect and people will know that they have to provide `HttpClient` in order to use the Effect. 

Check `getMetrics` for instance: https://github.com/blocksense-network/blocksense/blob/main/apps/e2e-tests/src/utils/metrics/metrics-fetcher.ts

MillaGenova (changes requested):

apps/docs.blocksense.network/components/FutureNodeForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+      e.preventDefault();
+
+      setError(null);
+
+      if (!email || loading || success) return;
+      if (!S.is(Email)(email)) return;
+
+      setLoading(true);
+      return Effect.runPromise(
+        Effect.gen(function* (_) {
```
DannyStoyanov, line 28: nit: we can omit the `_` as it is not necessary from certain typescript version (TypeScript (v5.5+)) - https://effect.website/docs/getting-started/using-generators/#adapter

apps/docs.blocksense.network/service/client.ts
RadoslavDimchev: As agreed, we’ll rewrite the API as we improve and decouple it, since we need to rework how the API object is obtained and how the endpoints are invoked.

PR #1548, opened 2025-07-30
[e2e-tests]: Refactor e2e test layers' infrastructure

DannyStoyanov:
### Summary

This PR introduces a structural refactor to the E2E test infrastructure, improving modularity and alignment with idiomatic `Effect` patterns. It also includes updates to the dependency list to support new capabilities via the `@effect` ecosystem.

---

#### Refactor: E2E Infrastructure

- Removed the following redundant service layer implementation:
  - `ProcessComposeLive`
  - `SequencerLive`
  - `RGLogCheckerLive`
- Introduced `static Live` with layer's implementation on:
  - `ProcessCompose`
  - `Sequencer`
- Removed `RGLogChecker` context/service entirely.
  - Replaced with a new helper `rgSearchForPattern` using `Command.make` from `@effect/platform-node`.
- Updated all test usages to use the new `Live` layers directly.

#### Refactor: Polling Logic with `Effect.retry`

- Replaced all usages of the custom `loopWhile` helper with idiomatic `Effect.retry`.

  - Applied to polling of `process-compose` status
  - Applied to `updates_to_networks` metric convergence

- This provides a cleaner and more declarative retry model using built-in scheduling semantics.

Reference: [https://effect.website/docs/error-management/retrying/#retry](https://effect.website/docs/error-management/retrying/#retry)

#### Test Runtime Behavior

- Updated all polling-based tests to use `it.live` to reflect their dependence on real-time external conditions and timers.

#### Helpers

- Added `rgSearchForPattern` to `helpers.ts`:
  - Replaces `execa`-based grep with Effect-native CLI execution.
  - Uses `@effect/platform-node` and `NodeContext`.

---

### Dependencies Installed

The following packages were added to `apps/e2e-tests/package.json`:

- `@effect/platform-node`
- `@effect/cluster`
- `@effect/rpc`
- `@effect/sql`
- `@effect/experimental`

These dependencies are also reflected in the updated `yarn.lock`.

### DIFF
--- a/apps/e2e-tests/src/process-compose/e2e.spec.ts
+++ b/apps/e2e-tests/src/process-compose/e2e.spec.ts
@@ -1,15 +1,14 @@
-import { Effect, pipe } from 'effect';
+import { Effect, pipe, Schedule } from 'effect';
 import { afterAll, beforeAll, describe, expect, it } from '@effect/vitest';
 import { deepStrictEqual } from 'assert';
 
-import { loopWhile } from '@blocksense/base-utils/async';
 import { getProcessComposeLogsFiles } from '@blocksense/base-utils/env';
 import { entriesOf, mapValuePromises, valuesOf } from '@blocksense/base-utils';
 import { AggregatedDataFeedStoreConsumer } from '@blocksense/contracts/viem';
 
 import { rgSearchPattern, parseProcessesStatus } from './helpers';
 import { expectedPCStatuses03 } from './expected';
-import type { ProcessComposeService, UpdatesToNetwork } from './types';
+import type { ProcessComposeService } from './types';
 import { ProcessCompose, Sequencer } from './types';
 
 describe.sequential('E2E Tests with process-compose', () => {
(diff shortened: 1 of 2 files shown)


apps/e2e-tests/src/process-compose/helpers.ts
```diff
@@ -1,10 +1,13 @@
...
+import { Effect, ParseResult, Schema as S } from 'effect';
+import { NodeContext } from '@effect/platform-node';
 import { $, execa } from 'execa';
 
 import { logMessage } from '../utils/logs';
 
 import { arrayToObject } from '@blocksense/base-utils/array-iter';
 import { rootDir } from '@blocksense/base-utils/env';
+import { Command } from '@effect/platform';
+import { RGLogCheckerError } from './types';
```
MillaGenova, line 10:
```suggestion
import { Effect, ParseResult, Schema as S } from 'effect';
import { NodeContext } from '@effect/platform-node';
import { Command } from '@effect/platform';
import { $, execa } from 'execa';

import { logMessage } from '../utils/logs';

import { arrayToObject } from '@blocksense/base-utils/array-iter';
import { rootDir } from '@blocksense/base-utils/env';

import { RGLogCheckerError } from './types';
```

MillaGenova (approved):

apps/e2e-tests/src/process-compose/helpers.ts
```diff
@@ -64,3 +67,27 @@ export function logTestEnvironmentInfo(
...
+}: {
+  caseInsensitive?: boolean;
+  file: string;
+  flags?: Array<string>;
+  pattern: string;
+}) => {
+  const args = ['--quiet'];
+  if (caseInsensitive) args.push('-i');
+  if (flags) flags.forEach(f => args.push(f));
+  args.push(pattern, file);
```
PetarKirov, line 85:
```suggestion
  const args = caseInsensitive
    ? ['--quiet', '-i', ...flags, pattern, file]
    : ['--quiet', ...flags, pattern, file];
```

apps/e2e-tests/src/process-compose/helpers.ts
```diff
@@ -64,3 +67,27 @@ export function logTestEnvironmentInfo(
...
+  caseInsensitive = true,
+  file,
+  flags = [],
+  pattern,
+}: {
+  caseInsensitive?: boolean;
+  file: string;
+  flags?: Array<string>;
+  pattern: string;
+}) => {
```
PetarKirov, line 81: Add return type to make the behaviour clear:

* If goal of this function is to only tell if a match has been found, the success type should be `boolean`
* If it you need the matches (say get all logs with "result from network") then perhaps it should return:
  ```ts
  type RipGrepMatch = {
    file: string;
    lineNumber: number;
    match: string; // Consider regex match or the whole line
  }[];
  ```

Perhaps there should be one generic function (that takes a list of flags) and some helpers that either return boolean or list of results on top.

EmilIvanichkovv (approved):

MillaGenova (approved):

PR #1550, opened 2025-07-31
fix(nix): Do not set `LD_LIBRARY_PATH` in the rust `devShell`

reo101:
- Leads to the breakage of all in-shell applications which originally
  link to (usually) older versions of the libraries in `LD_LIBRARY_PATH`
  (biggest offender being `nix` itself with `libcurl`)
- Set `LD_LIBRARY_PATH` for the sequencer and reporters manually in the
  `process-compose` configuration
  (they were previously inheriting it from the shell)


kkirkov (approved):
I had a problem with finding `/nix/store/v578vkzh0qhzczjvrzf64lqb2c74d5pk-curl-8.13.0/lib/libcurl.so.4 `
This PR fixes it. TY!

EmilIvanichkovv (approved):

PR #1561, opened 2025-08-05
chore(trigger-oracle): Set logging level to debug on progress of execute_wasm

HristoStaykov:


PetarKirov (approved):
This will make the logs easier to filter, so it's very nice to have 👍

PR #1576, opened 2025-08-12
feat(oracles): Add timeout when fetching prices

yordanmadzhunkov:
Currently a stop in Gate.io caused the whole crypto price oracle to stop working, dispute having available data from other exchanges. This PR adds the ability to handle this kind of situation


apps/oracles/stock-price-feeds/src/fetch_prices.rs
```diff
@@ -80,7 +81,7 @@ pub async fn get_prices(
         fetch::<FMPPriceFetcher>(&symbols.fmp, get_api_keys(capabilities, &["FMP_API_KEY"])),
     ]);
 
-    let fetched_provider_prices = fetch_all_prices(futures_set).await;
+    let fetched_provider_prices = fetch_all_prices(futures_set, &Duration::from_secs(15)).await;
```
HristoStaykov, line 84: Should we increase this timeout as well, the current voting period of all price feeds is 90 seconds, we could go for the same value as in cex-price-feeds might be?

PR #1580, opened 2025-08-13
feat(nix/shells): Add `oracle-development` devshell

EmilIvanichkovv:
It set the `LD_LIBRARY_PATH` env. Needed to run `just start-oracle <name>`

### DIFF
--- a/nix/shells/default.nix
+++ b/nix/shells/default.nix
@@ -39,6 +39,10 @@
           shellName = "Lint";
           extraImports = [ ];
         };
+        oracle-development = createShell {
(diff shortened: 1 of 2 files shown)


reo101:
In order to prevent the *fixed* issue of old `nix` users having their `nix` rendered unusable due to the newer `libcurl` in `LD_LIBRARY_PATH`, could we not add a `enterShell` entry for comparing the nix version of the running system with that in the `devShell`'s `nixpkgs` instance? Comparing `$(nix --version)` and `$(${lib.getExe pkgs.nixVersions.latest} --version)`?

PR #1582, opened 2025-08-13
feat(nix): Add option to use nix derivations as well as cargo artifacts in process-compose

PetarKirov:
This change makes the behavior of commit de588974fc4ffebcbbe0d42d57487aab063c102f
configurable.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -30,6 +30,8 @@ build-environment environment="all" use-local-cargo-result="0":
   #!/usr/bin/env bash
   set -euo pipefail
 
+  echo "Building process-compose environment: {{environment}}"
+
   if [[ "{{use-local-cargo-result}}" = 1 ]]; then
     FLAKE_ATTR_PATH=process-compose-environments.with-local-cargo-artifacts.{{environment}}
   else
@@ -47,8 +49,9 @@ build-environment environment="all" use-local-cargo-result="0":
   scripts/utils/collect-available-ports.sh "$DEST_DIR/available-ports"
 
   SRC_DIR=$(
-    nix build --impure -L --print-out-paths \
-      .#${FLAKE_ATTR_PATH}
+    nix build --no-warn-dirty --impure -L --print-out-paths \
+      .#${FLAKE_ATTR_PATH} \
+      2> >(grep -v '^Using saved setting for' >&2)
   )
 
   cp -rf --no-preserve=mode,ownership "$SRC_DIR"/. "$DEST_DIR"
--- a/Justfile
+++ b/Justfile
@@ -26,26 +26,42 @@ list-environments:
 
 [group('Working with process-compose environments')]
 [doc('Build process-compose artifacts for a specific environment or all environments')]
-build-environment environment="all":
+build-environment environment="all" use-local-cargo-result="0":
   #!/usr/bin/env bash
   set -euo pipefail
-  # Collect free ports that process-compose will use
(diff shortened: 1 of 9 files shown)


nix/shells/pkg-sets/rust.nix
```diff
@@ -16,7 +15,7 @@
 
   packages = self'.packages.blocksense-rs.buildInputs ++ [
     self'.legacyPackages.cargoWrapped
-    inputs'.nixpkgs-unstable.legacyPackages.fermyon-spin
```
reo101, line 19: Is this meant to be done in this commit/PR?

nix/test-environments/default.nix
```diff
@@ -16,44 +16,64 @@
...
-      allEnvironments = lib.pipe allEnvironmentNames [
-        (builtins.map (name: {
-          inherit name;
-          value = pkgs.runCommand "process-compose-${name}" { } ''
-            mkdir -p "$out"
-            ln -s ${
-              config.devenv.shells.${name}.process.managers.process-compose.configFile
-            } "$out/process-compose.yaml"
+      allEnvironments =
+        use-local-cargo-result:
```
reo101, line 20: Can we opt into passing an attrset like `{ local : bool }` here?

nix/test-environments/default.nix
```diff
@@ -16,44 +16,64 @@
...
 
-      allProcessComposeFiles = pkgs.runCommand "allProcessComposeFiles" { } ''
-        mkdir "$out"
-        (
-          set -x
-          ${lib.concatMapStringsSep "\n" (x: "cp -r ${x.value} \"$out/${x.name}\"") allEnvironments}
-        )
-      '';
+      allProcessComposeFiles =
+        use-local-cargo-result:
```
reo101, line 42: Same here

nix/test-environments/default.nix
```diff
@@ -16,44 +16,64 @@
...
-      devenv.shells = lib.genAttrs allEnvironmentNames (name: {
-        imports = [
-          self.nixosModules.blocksense-process-compose
-          ./${name}.nix
-        ];
-      });
+      devenv.shells =
+        let
+          shellCombinations =
+            allEnvironmentNames ++ (builtins.map (name: "${name}-use-local-cargo-result") allEnvironmentNames);
```
reo101, line 67: Can we opt into passing an attrset like `{ name : string, local : bool }` here too?

It should greatly simplify the iteration logic below

Justfile
```diff
@@ -38,15 +38,23 @@ build-environment environment="all":
     cp -rf --no-preserve=mode,ownership "$srcDir"/. {{process-compose-artifacts-dir}}
   else
     destDir="{{process-compose-artifacts-dir}}/{{environment}}"
-    srcDir=$(nix build --impure -L --json .#legacyPackages.{{system}}.process-compose-environments.{{environment}} | jq -r '.[0].outputs.out')
+    if [[ "${USE_LOCAL_CARGO_RESULTS:-0}" = 1 ]]; then
```
reo101, line 41: Could we abstract the only difference between these if branches (the `process-compose-environments{,-with-cargo-local}` into a variable (either a shell or Justfile one) and have only one `srcDir=...` line then

MartinNikov (approved):

EmilIvanichkovv (approved):
I am okay but maybe consider @reo101’s comments

EmilIvanichkovv (approved):

DannyStoyanov (approved):

PR #1584, opened 2025-08-18
[cost-calculation]: Update Cronos Testnet specific logic

kkirkov:


libs/ts/onchain-interactions/src/scripts/cost-calculations.ts
```diff
@@ -226,30 +226,30 @@ const fetchTransactionsForNetwork = async (
       );
       rawTransactions = response.data.result.records || [];
     } else if (network === 'cronos-testnet') {
-      let currentPage = 1;
-      let totalPages = 1;
+      let pageCounter = 1;
+      const numOfPages = 30n; //max 10000 blocks per page
```
Owliie, line 230: `totalPages` is a better name. Also why do you pin exactly 30?

libs/ts/onchain-interactions/src/scripts/cost-calculations.ts
```diff
@@ -226,30 +226,30 @@ const fetchTransactionsForNetwork = async (
...
+            `Found ${txFromPage.length} transactions on page ${pageCounter}/${numOfPages}`,
+          ),
+        );
         rawTransactions = rawTransactions.concat(txFromPage);
-        totalPages = page.data.pagination.totalPage;
-        currentPage += 1;
-      } while (currentPage <= totalPages);
+        currentStartBlock += 10000n;
+        pageCounter++;
+      } while (currentStartBlock < latestBlock);
```
Owliie, line 252: What happens if we call the api with an `endblock` which is greater than the latest?

Owliie (changes requested):

Owliie (approved):

PR #1585, opened 2025-08-18
Timing out proxy for oracle scripts

HristoStaykov:


apps/oracles/cex-price-feeds/spin.toml
```diff
@@ -4431,21 +4431,7 @@ data = '{"pair":{"base":"USDC","quote":"USD"},"decimals":8,"category":"Crypto","
...
-  "https://api.mexc.com",
-  "https://api.crypto.com",
-  "https://api.binance.us",
-  "https://api.gemini.com",
-  "https://api-pub.bitfinex.com",
-  "https://api.upbit.com",
-  "https://api.bitget.com",
-  "https://api.gateio.ws",
-  "https://www.okx.com",
+  "http://127.0.0.1:3000",
```
reo101, line 4434: Are we going to reintroducte this list of allowed outbound hosts in the new proxy somehow? We can start by *hardcoding* it in `Rust` directly (for now), leaving the possibility to abstract it away in a configuration file somewhere, similar to this one.

apps/oracles/cex-price-feeds/src/fetch_prices.rs
```diff
@@ -56,24 +56,24 @@ impl SymbolsData {
...
-        fetch::<CoinbasePriceFetcher>(&symbols.coinbase, None),
-        fetch::<CryptoComPriceFetcher>(&[], None),
+        fetch::<BinancePriceFetcher>(&symbols.binance, None, timeout_secs),
+        fetch::<BinanceUsPriceFetcher>(&symbols.binance_us, None, timeout_secs),
+        fetch::<BitfinexPriceFetcher>(&symbols.bitfinex, None, timeout_secs),
+        fetch::<BitgetPriceFetcher>(&[], None, timeout_secs),
+        fetch::<BybitPriceFetcher>(&[], None, timeout_secs),
+        fetch::<CoinbasePriceFetcher>(&symbols.coinbase, None, timeout_secs),
+        fetch::<CryptoComPriceFetcher>(&[], None, timeout_secs),
         // fetch::<GateIoPriceFetcher>(&[], None),
```
reo101, line 70: Should we not add the `timeout_secs` parameter here as well, just for completeness' sake?

```suggestion
        // fetch::<GateIoPriceFetcher>(&[], None, timeout_secs),
```

apps/oracles/cex-price-feeds/src/lib.rs
```diff
@@ -43,9 +43,10 @@ struct Data {
 async fn oracle_request(settings: Settings) -> Result<Payload> {
     println!("Starting oracle component");
 
+    let timeout_secs = settings.interval_time_in_seconds - 1;
```
reo101, line 46: Why the `- 1` here?

apps/trigger-oracle/src/main.rs
```diff
@@ -1,19 +1,206 @@
...
+                        tracing::info!("Starting timing out server on port 3000 ...");
+                        match HttpServer::new(|| App::new().service(timing_out_request))
+                            .bind("127.0.0.1:3000")
+                            .expect("Could not start timing out server on port 3000")
+                            .run()
+                            .await
+                        {
+                            Ok(v) => Ok(v),
+                            Err(e) => Err(anyhow::anyhow!(e.to_string())),
+                        }
```
reo101, line 189:
```suggestion
                        HttpServer::new(|| App::new().service(timing_out_request))
                            .bind("127.0.0.1:3000")
                            .expect("Could not start timing out server on port 3000")
                            .run()
                            .await
                            .context("Failed timing out server")
```

libs/sdk/src/http.rs
```diff
@@ -8,40 +8,58 @@ use url::Url;
...
-    base_url: &str,
+    forward_url: &str,
     params: Option<&[QueryParam<'_, '_>]>,
     headers: Option<&[HeaderParam<'_, '_>]>,
+    timeout_secs: u64,
 ) -> Result<Request> {
-    let url = match params {
-        Some(p) => Url::parse_with_params(base_url, p)?,
-        None => Url::parse(base_url)?,
+    let base_url = "http://127.0.0.1:3000";
```
reo101, line 25: Should this (host + port) stay hardcoded like this?

nix/modules/blocksense/backends/process-compose.nix
```diff
@@ -97,6 +97,7 @@ let
           };
           log_configuration = logsConfig;
           log_location = cfg.logsDir + "/reporter-${name}.log";
+          shutdown.signal = 9;
```
reo101, line 100: Why is this needed now?

nix/test-environments/example-setup-01.nix
```diff
@@ -199,21 +196,7 @@ in
...
-          "https://api.mexc.com"
-          "https://api.crypto.com"
-          "https://api.binance.us"
-          "https://api.gemini.com"
-          "https://api-pub.bitfinex.com"
-          "https://api.upbit.com"
-          "https://api.bitget.com"
-          "https://api.gateio.ws"
-          "https://www.okx.com"
+          "http://127.0.0.1:3000"
```
reo101, line 199: Should we not set this as the default for the `Nix` option at this point? Maybe remove the option altogether (or modify it's implementation to provide the `allowed-outbound-hosts` to the proxy somehow)?

libs/sdk/src/http.rs
HristoStaykov: We can introduce the configuration of the url in a following PR.

reo101 (approved):

libs/data_providers_sdk/src/price_data/fetchers/exchanges/gemini.rs
```diff
@@ -33,20 +33,25 @@ impl<'a> PricesFetcher<'a> for GeminiPriceFetcher<'a> {
...
 
             while let Some(result) = futures.next().await {
-                if let Ok((symbol, price_pint)) = result {
-                    prices.insert(symbol, price_pint);
+                match result {
+                    Ok((symbol, price_pint)) => {
+                        prices.insert(symbol, price_pint);
+                    }
+                    Err(err) => {
+                        anyhow::bail!("Error processing future in GeminiPriceFetcher {err:?}")
```
EmilIvanichkovv, line 53: Use `eprintln!` or another way to log. Do not bail, as if we have 10 assets = 10 requests, and one of them fails, but the rest succeed, we do not want to nuke all.

libs/data_providers_sdk/src/price_data/fetchers/exchanges/coinbase.rs
```diff
@@ -34,20 +34,25 @@ impl<'a> PricesFetcher<'a> for CoinbasePriceFetcher<'a> {
...
             while let Some(result) = futures.next().await {
-                if let Ok((symbol, price_pint)) = result {
-                    prices.insert(symbol, price_pint);
+                match result {
+                    Ok((symbol, price_pint)) => {
+                        prices.insert(symbol, price_pint);
+                    }
+                    Err(err) => {
+                        anyhow::bail!("Error processing future in CoinbasePriceFetcher {err:?}")
+                    }
```
EmilIvanichkovv, line 55: diffo

nix/modules/blocksense/backends/process-compose.nix
EmilIvanichkovv: I am wondering if this will affect the `e2e-tests`. Right now, if you start the test setup but terminate it with `CTRL-C` process-compose will not die. I hope this will have a positive effect there too

PR #1587, opened 2025-08-19
fix(providers): Log instead of panicking

Owliie:


MartinNikov (approved):

EmilIvanichkovv (approved):
Offline, we agreed that this is a hotfix of the hotfix.. and we need it now

apps/sequencer/src/providers/provider.rs
```diff
@@ -238,7 +238,7 @@ async fn load_data_from_chain(
                 provider.round_counters = round_counters;
             }
             Err(err) => {
-                panic!("Error when loading round counters for {network} = {err}");
+                error!("Error when loading round counters for {network} = {err}");
```
PetarKirov, line 241: I agree with this change, as a problem on any network will shutdown the sequencer on for all networks. That said, what would be the state of the round counters in case of error? Consider setting the round counters to 0, to make this explicit.

PetarKirov (approved):

PR #1588, opened 2025-08-19
feat(github): Add merge workflows

MartinNikov:
⛓️ Dependency: This PR depends on https://github.com/metacraft-labs/nixos-modules/pull/287 and must not be merged until that one is complete.

### DIFF
--- a/.github/workflows/merge-main-to-testnet.yml
+++ b/.github/workflows/merge-main-to-testnet.yml
@@ -1,0 +1,11 @@
+name: 'Merge Main to Testnet'
+on:
+  workflow_dispatch:
+
+jobs:
+  promote_to_testnet:
+    name: Merge Main into Testnet
+    uses: blocksense-network/nixos-modules/.github/workflows/reusable-merge.yml@main
+    with:
+      source_branch: 'main'
+      target_branch: 'testnet'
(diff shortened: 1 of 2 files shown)


.github/workflows/merge-main-to-testnet.yml
```diff
@@ -0,0 +1,11 @@
+name: 'Merge Main to Testnet'
```
reo101, line 1: We need to be very careful with not messing `main` and `mainnet` up, maybe consider renaming `main` (as a *net) to something like `gitmain` or `head`/`githead`?

.github/workflows/merge-main-to-testnet.yml
```diff
@@ -0,0 +1,11 @@
...
+on:
+  workflow_dispatch:
+
+jobs:
+  promote_to_testnet:
+    name: Merge Main into Testnet
+    uses: blocksense-network/nixos-modules/.github/workflows/reusable-merge.yml@main
```
reo101, line 8: Are we fine with pinning `@main` here?

.github/workflows/merge-main-to-testnet.yml
MartinNikov: Although we discussed changing `main` to `devnet`, I don't believe it's official yet. After the branch is renamed, I'll be committed to changing it here.

reo101 (approved):

PR #1591, opened 2025-08-21
Fix oracle scripts

Owliie:
* Make error handling in Coinbase and Gemini better: `bail!` if no prices fetched, log otherwise
* Don't log `payload.endpoint_url` in trigger oracle timing out proxy, but only the scheme and host of the url
* Optional use proxy for http_get_json & http_post_json in `sdk/http.rs`

### DIFF
--- a/apps/trigger-oracle/src/main.rs
+++ b/apps/trigger-oracle/src/main.rs
@@ -61,9 +61,22 @@ async fn timing_out_request(req: HttpRequest, payload: web::Payload) -> impl Res
                     break;
                 }
             }
+
             if !allowed {
-                tracing::error!("To allow requests, add 'allowed_outbound_hosts = [\"{}\"]' to the manifest component section.", payload.endpoint_url);
-                let err_msg = format!("Destination not allowed: {}", payload.endpoint_url);
+                let (uri_scheme, uri_host) = match (payload_uri.scheme_str(), payload_uri.host()) {
+                    (Some(uri_scheme), Some(uri_host)) => (uri_scheme, uri_host),
+                    _ => {
+                        let err_msg = format!(
+                            "Invalid URI scheme or host in payload: {}",
+                            payload.endpoint_url
+                        );
+                        tracing::error!(err_msg);
+                        return HttpResponse::BadRequest().body(err_msg);
+                    }
+                };
+
+                tracing::error!("To allow requests, add 'allowed_outbound_hosts = [\"{uri_scheme}://{uri_host}\"]' to the manifest component section.");
+                let err_msg = format!("Destination not allowed: {uri_scheme}://{uri_host}");
                 tracing::error!(err_msg);
                 return HttpResponse::BadRequest().body(err_msg);
             }
--- a/libs/data_providers_sdk/src/price_data/fetchers/exchanges/coinbase.rs
+++ b/libs/data_providers_sdk/src/price_data/fetchers/exchanges/coinbase.rs
@@ -51,9 +51,13 @@ impl<'a> PricesFetcher<'a> for CoinbasePriceFetcher<'a> {
                         prices.insert(symbol, price_pint);
                     }
                     Err(err) => {
-                        anyhow::bail!("Error processing future in CoinbasePriceFetcher {err:?}")
+                        eprintln!("Error processing future in CoinbasePriceFetcher {err:?}")
                     }
                 }
+            }
+
+            if prices.is_empty() {
+                anyhow::bail!("No prices fetched from Coinbase");
             }
 
             Ok(prices)
--- a/libs/data_providers_sdk/src/price_data/fetchers/exchanges/coinbase.rs
+++ b/libs/data_providers_sdk/src/price_data/fetchers/exchanges/coinbase.rs
@@ -70,7 +70,8 @@ pub async fn fetch_price_for_symbol(
     timeout_secs: u64,
 ) -> Result<(String, PricePoint)> {
     let url = format!("https://api.exchange.coinbase.com/products/{symbol}/ticker");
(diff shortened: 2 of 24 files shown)


libs/sdk/src/http.rs
```diff
@@ -20,32 +20,42 @@ pub fn prepare_get_request(
...
+                endpoint_url: forward_url_and_params.to_string(),
+                seconds: timeout_secs,
+                request_method: "GET".to_string(),
+                request_body: None,
+            };
 
-    req.body(serde_json::to_string(&payload).unwrap());
+            req.body(serde_json::to_string(&payload).unwrap());
+        }
+        _ => {
```
reo101, line 52: nit: Don't be so cheap :smile:
```suggestion
        None => {
```

apps/trigger-oracle/src/main.rs
```diff
@@ -62,8 +62,12 @@ async fn timing_out_request(req: HttpRequest, payload: web::Payload) -> impl Res
...
             }
             if !allowed {
-                tracing::error!("To allow requests, add 'allowed_outbound_hosts = [\"{}\"]' to the manifest component section.", payload.endpoint_url);
-                let err_msg = format!("Destination not allowed: {}", payload.endpoint_url);
+                tracing::error!("To allow requests, add 'allowed_outbound_hosts = [\"{}://{}\"]' to the manifest component section.", payload_uri.scheme_str().unwrap_or(""), payload_uri.host().unwrap_or(""));
+                let err_msg = format!(
+                    "Destination not allowed: {}://{}",
+                    payload_uri.scheme_str().unwrap_or(""),
+                    payload_uri.host().unwrap_or("")
+                );
```
reo101, line 70: These `unwrap_or`s are fishy, discussing in DMs

libs/sdk/src/http.rs
```diff
@@ -20,32 +20,42 @@ pub fn prepare_get_request(
...
+            req.method(Method::Get);
+            req.uri(forward_url_and_params.as_str());
+            req.header("Accepts", "application/json");
+            req.header("User-Agent", "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36");
```
reo101, line 56: Why? It was like that before #1585 too

libs/sdk/src/http.rs
```diff
@@ -76,32 +86,45 @@ where
...
+                request_method: "POST".to_string(),
+                request_body: Some(request_body),
+            };
+
+            req.body(serde_json::to_string(&payload).unwrap());
+        }
+        _ => {
+            req.uri(Url::parse(forward_url)?.as_str());
+            req.header("Content-Type", "application/json");
+            req.header("User-Agent", "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36");
```
reo101, line 115: ditto

libs/sdk/src/http.rs
Owliie: The idea is to preserve the old logic for non-`timing out proxy` requests

libs/sdk/src/http.rs
```diff
@@ -76,32 +86,45 @@ where
...
-    };
+    match timeout_secs {
+        Some(timeout_secs) => {
+            req.uri("http://127.0.0.1:3000");
 
-    req.body(serde_json::to_string(&payload).unwrap());
+            let payload = Params {
+                endpoint_url: forward_url.to_string(),
+                seconds: timeout_secs,
+                request_method: "POST".to_string(),
```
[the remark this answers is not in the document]
reo101: That's not concerning the same request, if I understand it correctly

libs/data_providers_sdk/src/price_data/fetchers/exchanges/coinbase.rs
```diff
@@ -51,11 +51,15 @@ impl<'a> PricesFetcher<'a> for CoinbasePriceFetcher<'a> {
                         prices.insert(symbol, price_pint);
                     }
                     Err(err) => {
-                        anyhow::bail!("Error processing future in CoinbasePriceFetcher {err:?}")
+                        eprintln!("Error processing future in CoinbasePriceFetcher {err:?}")
```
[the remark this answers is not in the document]
EmilIvanichkovv: Maybe just outdated feeds.
image


The last line shows we obtain some info after all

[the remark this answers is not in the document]
EmilIvanichkovv: It seems we need `User-agent` header after all
image

libs/sdk/src/http.rs
```diff
@@ -20,41 +20,51 @@ pub fn prepare_get_request(
...
-        endpoint_url: forward_url_and_params.to_string(),
-        seconds: timeout_secs,
-        request_method: "GET".to_string(),
-        request_body: None,
+    req.header("Accepts", "application/json");
+    req.header("User-Agent", "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36");
+
+    match timeout_secs {
+        Some(timeout_secs) => {
+            req.method(Method::Post);
```
EmilIvanichkovv, line 42: Is this correct?

Owliie: `timing out proxy` has only one handler which accepts POST. The reason why this is done so is to provide a `body` in the request (which was throwing spin errors when we tried with GET).

PR #1601, opened 2025-08-27
fix(trigger-oracle/timing_out_proxy): Fix signal handling - use custom signal handler instead of actix's

HristoStaykov:


EmilIvanichkovv (changes requested):
It seems this does not fix the problem with killing `just start-oracle <name>` with Ctr + C

Image

PR #1608, opened 2025-08-28
Refactor use state hash for sync rebased

HristoStaykov:
Initial version of using Merkle root over all call-data updates generated per network.


apps/sequencer/src/providers/eth_send_utils.rs
```diff
@@ -722,10 +778,18 @@ pub async fn eth_batch_send_to_contract(
...
 
     provider.update_history(&updates.updates);
+    let result = receipt.status().to_string();
+    if result == "true" {
+        //Transaction was successfully confirmed therefore we update the latest state hash
+        let root = next_calldata_merkle_tree.root();
+        provider.calldata_merkle_tree_frontier = next_calldata_merkle_tree;
+        provider.merkle_root_in_contract = None;
+        debug!("Successfully updated contract in network `{net}` block height {block_height} Merkle root {root:?}");
+    } // TODO: Reread round counters + latest state hash from contract
```
HristoStaykov: Comment not relevant to this code path. TODO: Remove.

PR #1631, opened 2025-09-02
[social-verification/ui]: Implement `LetsTalkEmail`component

RadoslavDimchev:
Implemented email content with React and Tailwind in our UI library and used it with `react-email` render function in the `sendEmail` endpoint, instead of plain HTML and CSS.

image


DannyStoyanov (approved):

PetarKirov (approved):
Looks like a solid refactoring. Some suggestions for the future:

* Extract parts of the content that may change as constants (e.g. link or stats)
* Implement unsubscribe logic
* Implement a wrapper `renderEmail` function that wraps its parameter in TailWind automatically

PR #1637, opened 2025-09-03
feat(decoder): Move decoders to a separate workspace

Owliie:
* Move `libs/ts/contracts/templates` to `libs/ts/decoders`
* Make EVM version optional (choose whether to use mcopy opcode)
* Add padding to encoded data up to a multiple of 32b
* Make lower case possible for struct names
* Prefix composite types with underscore to allow `data` name for fields
* Add example E2E contracts to decode stored data from ADFS


libs/ts/decoders/src/encode-packed/utils/helpers.ts
```diff
@@ -0,0 +1,46 @@
...
+    }
+  });
+
+  return containsDynamicData;
+};
+
+export const decoderImplementationsMap: Record<string, DecoderImplementations> =
+  {
+    cancun: {
+      generateDecoderPrimitiveLines: require('../helpers/primitiveFieldCancun')
```
PetarKirov: Make this a regular import at the top-level, as these files don't have any dependencies that we want to conditionally import

libs/ts/decoders/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@blocksense/decoders",
+  "scripts": {
+    "clean": "git clean -fdx -e .env",
+    "build": "hardhat compile && yarn build:typecheck",
+    "test": "hardhat test",
+    "test:paris": "EVM_VERSION=paris yarn hardhat test",
+    "ts": "yarn node --import tsx --no-deprecation",
```
PetarKirov: I think it's not needed, consider removing it

```suggestion
```

libs/ts/decoders/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "@nomicfoundation/hardhat-network-helpers": "^1.1.0",
+    "@nomicfoundation/hardhat-toolbox": "^5.0.0",
+    "@typechain/ethers-v6": "^0.5.1",
+    "@typechain/hardhat": "^9.1.0",
+    "@types/chai": "4",
+    "@types/ejs": "^3",
+    "@types/mocha": "^10.0.10",
+    "@types/node": "^24.3.1",
+    "chai": "4",
+    "dotenv": "^16.4.7",
```
PetarKirov:
```suggestion
```

Direnv handles loading of `.env` files for us [here](https://github.com/blocksense-network/blocksense/blob/7a0250fc0da849b12b7c47f78fae4e7d362f5a28/.envrc#L9).

libs/ts/contracts/tsconfig.json
```diff
@@ -10,6 +10,6 @@
     "outDir": "dist",
     "rootDir": "."
   },
-  "include": ["**/*.ts"],
+  "include": ["**/*.ts", "abis/**/*.json"],
```
PetarKirov: This change doesn't seem to be necessary:

```suggestion
  "include": ["**/*.ts"],
```

libs/ts/contracts/package.json
```diff
@@ -30,7 +31,13 @@
...
       "import": "./dist/esm/utils/calldata-decoder.mjs",
       "types": "./lib/utils/calldata-decoder.ts"
-    }
+    },
+    "./utils/*": {
+      "require": "./test/utils/*.ts",
+      "types": "./test/utils/*.ts"
+    },
+    "./contracts/*": "./contracts/*",
+    "./artifacts/*": "./artifacts/*"
```
PetarKirov: Check if those two are still needed.
```suggestion
```

PetarKirov (approved):

MartinNikov (approved):

PR #1642, opened 2025-09-04
[changelog-generator/list-feeds]: Add filtering by `feed id` and `feed name`

DannyStoyanov:
### [BSN-3199: [list-feeds command] Add optional --feed-id and --feed-name options to allow filtering](https://coda.io/d/_d6vM0kjfQP6#_tu4Kik5j/r3199&view=full)

This pull request enhances the filtering capabilities of the `listFeeds` command in the `apps/changelog-generator/src/commands/list-feeds.ts` file. The main improvement is the addition of new filter options, allowing users to filter feeds by their ID and name, alongside the existing filters.

**New filter options:**

* Added `feedId` and `feedName` as optional command-line arguments to the `listFeeds` command, enabling more granular filtering of feeds.

**Filtering logic updates:**

* Updated the filtering logic to support the new `feedId` and `feedName` options, so feeds can now be filtered by ID or name in addition to the existing filters.


DannyStoyanov:
@PetarKirov do you think we have to make the `feed-name` option partial match instead of exact match?

MillaGenova (approved):
LGTM

PR #1675, opened 2025-09-16
[contracts-config]: Deployment of LME-XCU price in USD per troy ounce

kkirkov:


config/evm_contracts_deployment_v2/pharos-testnet.json
```diff
@@ -9563,6 +9563,1800 @@
...
+          "10135",
+          "0x72F88509C53b939a0613c679a0F4768c0444d247"
+        ],
+        "salt": "0xe124d7cc79a19705865fa21b784ba187cd393559e960c0c071132cb60354d1a3",
+        "feedId": "10135",
+        "base": null,
+        "quote": null
+      },
+      "20000": {
+        "address": "0xedfa3ccA6C4A223f43c8781E29ad30f65C6A2D5D",
```
EmilIvanichkovv, line 11348: The address here is different. Is that expected?

kkirkov: Yes, on Pharos we are using our fork of Safe, because there isn't an official - so all contracts are with different addresses.

EmilIvanichkovv (approved):

PR #1679, opened 2025-09-17
[dev/read/adfs]: Add human readable value flag in `adfs` command 

MillaGenova:
This pull request adds support for human-readable output when reading ADFS data feeds, introduces a utility for formatting timestamps, and provides comprehensive tests for the new parsing logic. The changes improve the developer experience and usability of the ADFS feed reader command by allowing users to view parsed values and timestamps in a more accessible format.
### Command:
```
yarn start feeds read adfs --network ink-sepolia --feed-id 0 --human-readable
```

### Human-Readable Output for ADFS Feed Reader

* Added a `--human-readable` flag to the `adfs` command, enabling parsed and formatted display of feed data values and timestamps instead of raw hex strings.
* Integrated logic to parse ADFS data and display value, value hex, timestamp, timestamp hex, and formatted timestamp when the flag is used, both in single and latest data modes. [[1]](diffhunk://#diff-634a2b82996de61b38d2e5f6a3eb1c4666871ee3af98c0cd20d64fbccc1b0515L94-R158) [[2]](diffhunk://#diff-634a2b82996de61b38d2e5f6a3eb1c4666871ee3af98c0cd20d64fbccc1b0515R185-R200)

### Utilities and Parsing

* Implemented the `parseADFSData` function to extract and format value and timestamp fields from ADFS hex data, handling both prefixed and non-prefixed hex strings as well as error cases.
* Added the `formatTimestamp` utility with options for formatting timestamps as human-readable strings, including automatic detection of milliseconds vs seconds, and optional display of raw values and units.

### Testing 

* Added unit tests for the `parseADFSData` function to verify correct parsing of value and timestamp fields, handling of short/invalid hex, and compatibility with hex strings lacking a `0x` prefix.

[BSN-3434: Add Human-Readable Value Flag to read-adfs ](https://coda.io/d/_d6vM0kjfQP6#_tu4Kik5j/r3434&view=full)


apps/dev/src/commands/feeds/read/adfs.ts
```diff
@@ -9,9 +9,43 @@ import {
...
 import { listEvmNetworks, readEvmDeployment } from '@blocksense/config-types';
 import { AggregatedDataFeedStoreConsumer } from '@blocksense/contracts/viem';
+import { formatTimestamp } from '../../utils';
 
 const availableNetworks = await listEvmNetworks();
 
+export const parseADFSData = (hexData: string) => {
+  const cleanHex = hexData.startsWith('0x') ? hexData.slice(2) : hexData;
+  const valuePart = cleanHex.slice(0, 48);
+  const valueHex = '0x' + valuePart.padEnd(64, '0');
```
EmilIvanichkovv, line 19: Setting this padding is incorrect.

The value is only in the first 48 bytes. So adding this padding makes a different value when you convert it to `bigint`
```suggestion
  const valueHex = '0x'+ cleanHex.slice(0, 48);
```

apps/dev/src/commands/feeds/read/adfs.ts
```diff
@@ -91,10 +139,23 @@ export const adfs = Command.make(
...
-          const data = yield* Effect.tryPromise(() =>
+          const single = yield* Effect.tryPromise(() =>
             consumer.getSingleDataAtIndex(feed, index.value),
           );
-          rows.push(['Data', data]);
+          if (isHumanReadable && !hasSlice && !multi) {
+            const parsed = parseADFSData(single);
+            rows.push(['Data (Raw)', parsed.rawHex]);
+            rows.push(['Value', parsed.value]);
+            rows.push(['Value (Hex)', parsed.valueHex]);
```
EmilIvanichkovv, line 149: no need to show `Value (Hex)` in the table

apps/dev/src/commands/feeds/read/adfs.ts
```diff
@@ -91,10 +139,23 @@ export const adfs = Command.make(
...
+            if (parsed.timestamp) {
+              rows.push(['Timestamp', parsed.timestamp]);
+              rows.push(['Timestamp (Hex)', parsed.timestampHex ?? '']);
```
EmilIvanichkovv, line 152: no need to show `Timestamp (Hex)` in the table

apps/dev/src/commands/feeds/read/adfs.ts
```diff
@@ -121,7 +182,20 @@ export const adfs = Command.make(
             consumer.getLatestSingleDataAndIndex(feed),
           );
           rows.push(['Index', index.toString()]);
-          rows.push(['Data', data as string]);
+          if (isHumanReadable && !hasSlice && !multi) {
+            const parsed = parseADFSData(data as string);
+            rows.push(['Data (Raw)', parsed.rawHex]);
+            rows.push(['Value', parsed.value]);
+            rows.push(['Value (Hex)', parsed.valueHex]);
```
EmilIvanichkovv, line 189: ditto

apps/dev/src/commands/feeds/read/adfs.ts
```diff
@@ -121,7 +182,20 @@ export const adfs = Command.make(
...
+            if (parsed.timestamp) {
+              rows.push(['Timestamp', parsed.timestamp]);
+              rows.push(['Timestamp (Hex)', parsed.timestampHex ?? '']);
```
EmilIvanichkovv, line 192: ditto

apps/dev/src/commands/feeds/read/adfs.ts
```diff
@@ -9,9 +9,43 @@ import {
 import { renderTui, drawTable } from '@blocksense/base-utils/tty';
 import { listEvmNetworks, readEvmDeployment } from '@blocksense/config-types';
 import { AggregatedDataFeedStoreConsumer } from '@blocksense/contracts/viem';
+import { formatTimestamp } from '../../utils';
 
 const availableNetworks = await listEvmNetworks();
 
+export const parseADFSData = (hexData: string) => {
```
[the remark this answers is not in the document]
EmilIvanichkovv: Also rename to `formatNumericalValue`

Here, this specific name is more valuable because the info stored in the ADFS contract can be more complex than just numbers. At this point, we do not have cases different than numerical data, but in the near future, there might be. When this happens, we will reconsider the `human-readable` flag

apps/dev/src/commands/feeds/read/adfs.ts
```diff
@@ -9,9 +9,43 @@ import {
...
+  const cleanHex = hexData.startsWith('0x') ? hexData.slice(2) : hexData;
+  const valuePart = cleanHex.slice(0, 48);
+  const valueHex = '0x' + valuePart.padEnd(64, '0');
+  let value: bigint = 0n;
+  try {
+    value = BigInt(valueHex);
+  } catch {
+    // ignore
+  }
```
EmilIvanichkovv, line 25: As we discussed, we only use this function for numerical data. Therefore if there is problem with the parsing, throwing here is fine as this might show incorrect usage.
```suggestion
  const value = BigInt(valueHex);
```

apps/dev/src/commands/utils.ts
EmilIvanichkovv, line 1: All the logic here can be replaced with the [`formatTimestamp`](https://github.com/blocksense-network/blocksense/blob/0589cbecfd1c95f72f78174b69d8c56f928d2153/apps/dev/src/commands/feeds/read/cl-adapter.ts#L17) from `apps/dev/src/commands/feeds/read/cl-adapter.ts`. There is no need to have such a complex implementation.

My suggestion is:
- move [`formatTimestamp`](https://github.com/blocksense-network/blocksense/blob/0589cbecfd1c95f72f78174b69d8c56f928d2153/apps/dev/src/commands/feeds/read/cl-adapter.ts#L17) to the `utils.ts` module
- convert it to a named function
- reuse it in both `adfs.ts` & `cl-adapter.ts`

apps/dev/src/commands/feeds/read/adfs.ts
```diff
@@ -9,9 +9,43 @@ import {
...
+  if (cleanHex.length >= 64) {
+    timestampHex = '0x' + cleanHex.slice(48, 64);
+    try {
+      timestamp = BigInt(timestampHex);
+    } catch {
+      timestamp = null;
+    }
+  }
+  const formattedTimestamp =
+    timestamp !== null ? formatTimestamp(timestamp) : null;
```
EmilIvanichkovv, line 37: Similarly, for the timestamp, we know it is in the data and it is the rest of the bytes, so we do not need complicated checks and parsing. Just:

```suggestion
  const timestampHex = '0x' + cleanHex.slice(48);
  const timestamp = BigInt(timestampHex);
  const formattedTimestamp = formatTimestamp(timestamp);
```

apps/dev/src/commands/feeds/read/adfs.ts
```diff
@@ -9,9 +9,43 @@ import {
...
+
+  return {
+    rawHex: hexData,
+    value: value.toString(),
+    valueHex,
+    timestamp: timestamp ? timestamp.toString() : null,
+    timestampHex,
+    formattedTimestamp,
+  } as const;
```
EmilIvanichkovv, line 46: In comments bellow I suggest we do not visualize hex representation of `value` and `timestamp` as it is present in the `rawHex` already. So you can simplify the return object.
```suggestion
  return {
    rawHex: hexData,
    value: value.toString(),
    timestamp: formattedTimestamp,
  };
```

apps/dev/src/commands/feeds/read/adfs.ts
```diff
@@ -91,10 +139,23 @@ export const adfs = Command.make(
...
+          if (isHumanReadable && !hasSlice && !multi) {
+            const parsed = parseADFSData(single);
+            rows.push(['Data (Raw)', parsed.rawHex]);
+            rows.push(['Value', parsed.value]);
+            rows.push(['Value (Hex)', parsed.valueHex]);
+            if (parsed.timestamp) {
+              rows.push(['Timestamp', parsed.timestamp]);
+              rows.push(['Timestamp (Hex)', parsed.timestampHex ?? '']);
+              if (parsed.formattedTimestamp)
+                rows.push(['Timestamp (Formatted)', parsed.formattedTimestamp]);
```
EmilIvanichkovv, line 154: With all the changes above the logic of the table rows is that simple:
```suggestion
            rows.push(['Data (Raw)', parsed.rawHex]);
            rows.push(['Value', parsed.value]);
            rows.push(['Timestamp', parsed.timestamp]);
```

apps/dev/src/commands/feeds/read/adfs.ts
```diff
@@ -121,7 +182,20 @@ export const adfs = Command.make(
...
+          if (isHumanReadable && !hasSlice && !multi) {
+            const parsed = parseADFSData(data as string);
+            rows.push(['Data (Raw)', parsed.rawHex]);
+            rows.push(['Value', parsed.value]);
+            rows.push(['Value (Hex)', parsed.valueHex]);
+            if (parsed.timestamp) {
+              rows.push(['Timestamp', parsed.timestamp]);
+              rows.push(['Timestamp (Hex)', parsed.timestampHex ?? '']);
+              if (parsed.formattedTimestamp)
+                rows.push(['Timestamp (Formatted)', parsed.formattedTimestamp]);
```
EmilIvanichkovv, line 194: Ditto
```suggestion
            rows.push(['Data (Raw)', parsed.rawHex]);
            rows.push(['Value', parsed.value]);
            rows.push(['Timestamp', parsed.timestamp]);
```

apps/dev/test/parseADFSData.test.ts
EmilIvanichkovv, line 1: You will need to update the tests accordingly

apps/dev/src/commands/feeds/read/adfs.ts
```diff
@@ -9,9 +9,43 @@ import {
 import { renderTui, drawTable } from '@blocksense/base-utils/tty';
 import { listEvmNetworks, readEvmDeployment } from '@blocksense/config-types';
 import { AggregatedDataFeedStoreConsumer } from '@blocksense/contracts/viem';
+import { formatTimestamp } from '../../utils';
 
 const availableNetworks = await listEvmNetworks();
 
+export const parseADFSData = (hexData: string) => {
+  const cleanHex = hexData.startsWith('0x') ? hexData.slice(2) : hexData;
```
EmilIvanichkovv, line 17: We can add a check if the passed data is of the correct length and can be converted to a numeric value and a timestamp.
```suggestion
  const cleanHex = hexData.startsWith('0x') ? hexData.slice(2) : hexData;
  if (cleanHex.length !== 64) {
    throw new Error(
      'Unexpected ADFS data length, expected 32 bytes (64 hex characters), got ' +
        cleanHex.length,
    );
  }
```

EmilIvanichkovv (changes requested):

DannyStoyanov (approved):

PR #1682, opened 2025-09-18
[rollup]: Replace `chalk-template` with manual ANSI

RadoslavDimchev:
Replace `chalk-template` with inline ANSI sequences in `apps/rollup/bundle.mjs` and remove the dependency. Keeps the same bold/underline/red styling.

image


apps/rollup/bundle.mjs
EmilIvanichkovv, line 1: Can we export color/bold/reset encodings as constants.
This will improve readability

EmilIvanichkovv (changes requested):

apps/rollup/bundle.mjs
EmilIvanichkovv: They can live in `libs/ts/base-utils/src/tty.ts`

apps/rollup/package.json
```diff
@@ -11,12 +11,12 @@
     "rollup:build-workspace": "yarn rollup-bundle $INIT_CWD"
   },
   "dependencies": {
+    "@blocksense/base-utils": "workspace:*",
```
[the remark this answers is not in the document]
EmilIvanichkovv: Yep CI confirms.. You can not use `@blocksense/bese-utils` in `rollup`.

Just stick with constants for the colors you need in this workspace.

MillaGenova (approved):

PR #1700, opened 2025-09-23
[e2e-tests]: Create sandboxed anvil instance for e2e tests

DannyStoyanov:
### Summary

- **refactor(nix/modules):**
  - Extracted `auto-impersonate` into a standalone option for flexibility.  
  - Renamed some variable in `e2e-tests.spec.ts` for better readability

- **feat(nix/modules):**
  - **Added support for running `anvil` in **offline-start** mode**.  
  - Introduced new options:  
    - `--fork-chain-id`  
    - `--fork-block-number`  
    - `--state`  
    - `--state-interval`  

- **feat(e2e-tests):**
  - Updated E2E tests to read initial round data from a local `anvil` instance instead of the remote network. 

****
## [BSN-3419: Tune feed round in e2e tests](https://coda.io/d/_d6vM0kjfQP6#_tu4Kik5j/r3419&view=full)
Fixes an issue where the round counter could go beyond the max round counter (8192) and getting such updates(historic data) could result in failing tests.

### Notes
- Eliminates e2e tests flakyness caused by the fact that we read initial round data from the network while the sequencer could potentially publish updates to the local `anvil` instance. Then, when comparing the round data from the initial setup we hit differences.

### DIFF
--- a/apps/e2e-tests/src/process-compose/e2e.spec.ts
+++ b/apps/e2e-tests/src/process-compose/e2e.spec.ts
@@ -216,7 +216,7 @@ describe.sequential('E2E Tests with process-compose', () => {
       // Info is fetched for specific round - the initial round of the feed
       // + number of updates that happened while the local sequencer was running
       // modulo the maximum number of history elements per feed
-      const currentFeedsInfo = yield* getDataFeedsInfoFromNetwork(
+      const feedInfoAfterUpdates = yield* getDataFeedsInfoFromNetwork(
         updatedFeedIds,
         contractAddress,
         url,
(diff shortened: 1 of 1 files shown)


MillaGenova (approved):

apps/e2e-tests/src/process-compose/e2e.spec.ts
```diff
@@ -249,9 +241,14 @@ describe.sequential('E2E Tests with process-compose', () => {
...
+        const actualRound =
+          (MAX_HISTORY_ELEMENTS_PER_FEED +
+            (round - initialFeedsInfo[id].round) -
+            1) %
+          MAX_HISTORY_ELEMENTS_PER_FEED;
         const actualData = historyData.aggregate_history[id].find(
-          feed => feed.update_number === round - initialFeedsInfo[id].round - 1,
-        );
+          feed => feed.update_number === actualRound,
+        ); // here the round counter is >= 2 (from previous test) so it is safe to subtract 1 from the update number
```
EmilIvanichkovv, line 251:
```suggestion
        const actualData = historyData.aggregate_history[id].find(
          // In history data, the indexing of updates starts from 0.
          // Hence, we need to subtract 1 from the number of updates
          feed => feed.update_number === updatesToNetworks[network][id] - 1,
        );
```

EmilIvanichkovv: At this point, we have the following:
- Some updates have happened.
- At `currentFeedsInfo`, we have data for rounds = initial round + num of updates according to the sequencer.
- We need to compare updated on-chain data with what the sequencer thinks has happened ( values from `historyData` ). So we can index `historyData` with the number of updates according to the sequencer - 1 ( as indexing in history data starts from 0 )

We do not need the complicated math that calculates `actualRound` ( which in fact is the num of updates, not a round value )

EmilIvanichkovv: We can add one more check - one that confirms that the sequencer does not overwrite the latest round.
```ts
        const expectedNumberOfUpdates =
          (MAX_HISTORY_ELEMENTS_PER_FEED +
            (round - initialFeedsInfo[id].round)) %
          MAX_HISTORY_ELEMENTS_PER_FEED;

        expect(expectedNumberOfUpdates).toEqual(updatesToNetworks[network][id]);
```
In `expectedNumberOfUpdates`, we calculate how many updates have happened after collecting the initial feed info. When we compare it with the number of updates according to the sequencer we can confirm the sequencer has proceeded from the last update and has not rewritten it (known previous bug )

apps/e2e-tests/src/process-compose/e2e.spec.ts
```diff
@@ -240,7 +233,6 @@ describe.sequential('E2E Tests with process-compose', () => {
       // Make sure that the feeds info is updated
       for (const [id, data] of entriesOf(currentFeedsInfo)) {
         const { round, value } = data;
```
EmilIvanichkovv, line 235: Maybe it will help if we rename some of the vars. I think it might improve readability
  - `currentFeedsInfo` -> `feedInfoAfterUpdates`
  - `round` -> `roundAfterUpdates`
  - `value` -> `valueAfterUpdates`
  - `actualData` -> `historyData`

MillaGenova (approved):

PR #1702, opened 2025-09-23
Mod ring buffer indices before storage

Owliie:
Currently, the sequencer can write whatever number that fits in 2b in the ring buffer table per feed id. This PR introduces modulo operation to ensure that the number written is not greater than `2^13 - 1`. Otherwise, when the number is above this maximum the client reading from the contract actually reads from another feed's storage.

### DIFF
--- a/libs/ts/contracts/contracts/AggregatedDataFeedStore.sol
+++ b/libs/ts/contracts/contracts/AggregatedDataFeedStore.sol
@@ -416,7 +416,17 @@ contract AggregatedDataFeedStore {
 
           // update pointer to start start of index data
           pointer := add(pointer, add(indexLength, 1))
-          sstore(add(RING_BUFFER_TABLE_ADDRESS, slot), calldataload(pointer))
+          sstore(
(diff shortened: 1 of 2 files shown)


libs/ts/contracts/contracts/AggregatedDataFeedStore.sol
```diff
@@ -416,7 +416,13 @@ contract AggregatedDataFeedStore {
 
           // update pointer to start start of index data
           pointer := add(pointer, add(indexLength, 1))
-          sstore(add(RING_BUFFER_TABLE_ADDRESS, slot), calldataload(pointer))
+          sstore(
+            add(RING_BUFFER_TABLE_ADDRESS, slot),
+            and(
+              calldataload(pointer),
+              0x1fff1fff1fff1fff1fff1fff1fff1fff1fff1fff1fff1fff1fff1fff1fff1fff
```
PetarKirov, line 423: Please add a comment explaining why we're masking the ring buffer table entry

PR #1704, opened 2025-09-24
fix(sequencer/eth_send_utils): Use both eth_getPriorityFee and eth_maxPriorityFeePerGas to decide on using legacy transactions

HristoStaykov:
This PR introduces a call on the provider to get the "eth_getPriorityFee". When this field is not available, i.e. the provider returns an error, the network is considered **Legacy** and we proceed with `gas price` only.


Owliie (commented):
Leave a detailed comment so that we know why this value is never used and why we need to fetch it

PR #1727, opened 2025-10-02
fix(sequencer/eth_send_utils): Fix to add `type` field if missing - needed by alloy parsing of tx receipt

HristoStaykov:
### PR Description

This PR improves the logic for retrieving Ethereum transaction receipts, with two main changes:

1. **Receipt `type` field handling**

   * Receipts returned by the Ontology Ethereum network are missing the `type` field, which breaks downstream code expecting it.
   * We now check whether the `type` field exists, and if not, insert it ourselves with `"0x0"` value, which represents legacy type.
   * This ensures consistent receipt structures across networks.

2. **Simplified receipt retrieval mechanism**

   * Refactored and streamlined the code used to fetch transaction receipts.
   * Reduced unnecessary complexity, making the flow easier to follow and maintain.

3. **Create configuration parameters for**

   * receipt_polling_back_off_period_ms
   *  transaction_retry_back_off_ms

### Impact

* Transaction receipts are now guaranteed to include a `type` field.
* Receipt fetching logic is simpler, more reliable, and easier to maintain.

### DIFF
--- a/apps/sequencer/src/providers/eth_send_utils.rs
+++ b/apps/sequencer/src/providers/eth_send_utils.rs
@@ -1262,10 +1262,16 @@ mod tests {
     use blocksense_config::{AllFeedsConfig, PublishCriteria};
     use blocksense_data_feeds::feeds_processing::VotedFeedUpdate;
 
+    use alloy::network::EthereumWallet;
+    use alloy::signers::local::PrivateKeySigner;
     use blocksense_utils::test_env::get_test_private_key_path;
     use rdkafka::message::ToBytes;
     use regex::Regex;
+    use std::fs;
     use std::str::FromStr;
+    // New imports for receipt parsing / provider construction
+    use alloy::primitives::address;
+    use alloy::providers::ProviderBuilder;
 
     fn extract_address(message: &str) -> Option<String> {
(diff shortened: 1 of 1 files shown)


Owliie (changes requested):

apps/sequencer/src/providers/eth_send_utils.rs
```diff
@@ -773,6 +730,80 @@ pub async fn eth_batch_send_to_contract(
...
+                        debug!("Successfully got receipt from RPC in network `{net}` block height {block_height} and address {sender_address} tx_hash = {tx_hash} receipt = {v:?}");
+                        return Ok(v);
+                    }
+                    Ok(None) => {
+                        let deadline =
+                            start_time_point + Duration::from_secs(transaction_retry_timeout_secs);
+                        if Instant::now() > deadline {
+                            eyre::bail!("Get tx_receipt returned None and deadline for getting it elapsed in network `{net}` block height {block_height}");
+                        }
+                        await_time(1000).await;
```
Owliie, line 761: I think this is too little time for polling. Maybe it should be a bigger number. Otherwise, we could flood the RPC with requests and get rate-limited. What do you think?

reo101 (approved):

